PR #10, opened 2024-07-26
feat(search-clear): add search field clear primitive

Mattewn99:
## PR Checklist

- [X] The commit message follows our guidelines: https://github.com/ng-primitives/ng-primitives/blob/main/CONTRIBUTING.md#-commit-message-guidelines

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Right now, after clicking on "clear", the button is still visible

- [ ] Yes
- [X] No


Mattewn99:
@ashley-hunter does it maybe make sense to move all the logic about the clear button (I mean also the visibility) inside the directive that at this point could be a `@Component`?

ashley-hunter:
Thank you very much for the contribution @Mattewn99. 

It is interesting that you brought this up as I was recently thinking about this exact thing myself. Ideally the optimal scenario is one that involves the least effort on the consumers part - and what I had thought might be the best solution was to introduce an `NgpSearchFieldClear` primitive that a user could attach to the clear button and it would handle the clearing for them on click. 

This would avoid the need for any JavaScript to be written on the consumers end at all - they wouldn't need to programmatically change state (like they currently do), nor would they need to use ViewChild and call a method on the directive instance. I would be interested in hearing your thoughts on this approach! Thanks again!

Mattewn99:
I also didn't like to always have to write the js part for clearing the input. I like the approach of creating a new `NgpSearchFieldClear` directive. I can try an implementation on this PR. Thanks a lot!

ashley-hunter:
Perfect! Thank you! I believe there should be some guidance on creating a new primitive directive in the contributing guide. If anything isn't clear or you need any help just reach out.

Mattewn99:
@ashley-hunter I pushed an implementation of the new primitive. I'm not sure about the styling but I thought it'd be nice to set it in the primitive so the user doesn't need to add it in the consumer. About the docs would you prefer to have a separate page for this primitive? I added just a small example in the `search-field.example` since I think this is going to be more or less the only use case for this primitive. What do you think? Thanks!

ashley-hunter:
Thank you! Yes, so I think we want primitives to have no styles associated with them - styles would always be defined by the user to give them flexibility and to align with our other components and the idea of being a "Headless" library (in otherwords, no styles). 

We can add a data attribute to the clear button to reflect the empty state of the input field (we do this on the search field already) and the user can show or hide the button however they like - most likely with `display: none`, but for things like animations or transitions this gives most flexibility.

I think we could move the directive into the existing search field folder rather than needing one specific for the clear directive. Apart from that looks great! If you need any help with the above notes, just let me know, Im happy to help get this over the finish line any way you need 😊

Mattewn99:
@ashley-hunter makes sense. I added a `data-empty` attribute 👍🏻

Mattewn99:
Maybe @ashley-hunter it'd be nice to have a slack channel or something similar for faster communication and news on the next features 🤩

ashley-hunter:
Thank you! Yes that's a good idea! It's not my area of expertise, I know I have some groups use Discord - but if there is a good reason to use a specific app I'm open to suggestions!

Mattewn99:
I think discord is also a good option!

ashley-hunter:
Great! Good suggestion, I will set this up tonight 😊

ashley-hunter:
Thank you for your contribution! All looks good!

I have setup a discord server here: https://discord.gg/NTdjc5r3gC, any issues accessing it let me know!

ashley-hunter:
@allcontributors

ashley-hunter:
@allcontributors please add @Mattewn99 for code and doc

ashley-hunter:
@allcontributors please add @Mattewn99 for code and doc

PR #57, opened 2024-09-08
fix: focus management

otodockal:
## PR Checklist

- [ ] The commit message follows our guidelines: https://github.com/ng-primitives/ng-primitives/blob/main/CONTRIBUTING.md#-commit-message-guidelines

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

Closes #

To fully support dialog focus management, we need to allow program origin in NgpFocusVisible.
+ fix accordion styling

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/ng-primitives/dialog/src/dialog-trigger/dialog-trigger.directive.ts
+++ b/packages/ng-primitives/dialog/src/dialog-trigger/dialog-trigger.directive.ts
@@ -16,6 +16,7 @@ import {
   TemplateRef,
   ViewContainerRef,
 } from '@angular/core';
+import { NgpFocusVisible } from 'ng-primitives/interactions';
 import { NgpDialogRef } from '../dialog/dialog-ref';
 import { NgpDialogContext, NgpDialogManager } from '../dialog/dialog.service';
 import { NgpDialogTriggerToken } from './dialog-trigger.token';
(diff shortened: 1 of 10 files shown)


packages/ng-primitives/dialog/src/dialog-trigger/dialog-trigger.directive.ts
```diff
@@ -24,6 +25,7 @@ import { NgpDialogTriggerToken } from './dialog-trigger.token';
   standalone: true,
   selector: '[ngpDialogTrigger]',
   exportAs: 'ngpDialogTrigger',
+  hostDirectives: [NgpFocusVisible],
```
ashley-hunter: Unfortunately I don't believe we can have this here, for example, if the trigger is an `ngpButton` then `NgpFocusVisible` would be applied by both `ngpButton` and `ngpDialogTrigger` which would throw an issue.
I guess the solution would be to use `ngpButton` on the trigger in the example.

ashley-hunter (changes requested):
Change looks good, just had one small suggestion. Thank you!

packages/ng-primitives/dialog/src/dialog-trigger/dialog-trigger.directive.ts
otodockal: You're right. I'll send an update. So, the same problem is with ngpAccordionTrigger AFIAK.

ashley-hunter: Well spotted, that is a good point, the same change should probably be made for it! Thank you!

otodockal: fixed.

ashley-hunter (approved):
Great work! Thank you!

PR #162, opened 2025-04-06
Migrate state system into new project

MGREMY:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes
- [x] Other... Please describe:

I created a new package for state named `ng-primitives-state` in order to publish it separately from ng-primitive. Could be usefull to have a State system in angular for many people.

## Issue

Closes #

- [x] Yes
- [ ] No

The secondary entry-point for `ng-primitives/state` has been changed to `ng-primitives-state/state`


ashley-hunter (changes requested):
Hey! Thank you so much for the contribution! I have a few thoughts on the package name and package.

I think we should probably name the package `@ng-primitives/state`, I "own" the `@ng-primitives` npm scope so I had been planning on adding any additional packages under that scope.

I think we can also just put it under the "packages/state" path.

I think we also need to update the ng-primitives package to have a dependency on the state package.

And I think it would be good if we retain `ng-primitives/state` and that file would just contain:

```ts
export * from '@ng-primitives/state';
```

This way people don't need to know that state is part of a separate package if they are already using ng-primitives, but we have that other package for those wanting to use it standalone.

If anything is unclear just let me know, or I'm happy to help with any of these suggestions if needed. 

Thanks again! Nice work! Will be a great addition!

ashley-hunter (approved):
Great work, thanks again for the contribution!

MGREMY:
> Great work, thanks again for the contribution!

My pleasure ! :)

PR #242, opened 2025-05-15
feat(file-upload): filter dropped files by accepted fileTypes and lim…

marcjulian:
…it to one when multiple is false

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

`NgpFileUpload` / `NgpFileDropzone` ignore the inputs `multiple` and `fileTypes` for the `drop` event listener.

`NgpFileUpload` filters files from the drop event to match the `fileTypes` and only returns one file when `multiple` is false.

- [x] Yes
- [ ] No

This might be a breaking change for someone that relied on the previous behavior for the drop event.

### DIFF
--- a/packages/ng-primitives/file-upload/src/file-dropzone/file-drop-filter.ts
+++ b/packages/ng-primitives/file-upload/src/file-dropzone/file-drop-filter.ts
@@ -10,11 +10,27 @@ export function fileDropFilter(
   return limitedFiles.length > 0 ? filesToFileList(limitedFiles) : null;
 }
 
-function isFileTypeAccepted(file: File, acceptedTypes: string[] | undefined) {
+export function isFileTypeAccepted(file: File, acceptedTypes: string[] | undefined) {
   // allow all file types if no types are specified
   if (!acceptedTypes || acceptedTypes.length === 0) return true;
 
-  return acceptedTypes.some(acceptedType => file.type.match(acceptedType));
+  const mimeType = file.type;
+  const fileName = file.name.toLowerCase();
+
+  return acceptedTypes.some(type => {
+    type = type.toLowerCase();
+
+    if (type.startsWith('.')) {
+      return fileName.endsWith(type);
+    }
+
+    if (type.endsWith('/*')) {
+      const baseType = type.replace('/*', '');
+      return mimeType.startsWith(baseType);
+    }
+
+    return mimeType === type;
+  });
 }
 
 function filesToFileList(files: File[]): FileList {
--- a/packages/ng-primitives/file-upload/src/file-dropzone/file-drop-filter.ts
+++ b/packages/ng-primitives/file-upload/src/file-dropzone/file-drop-filter.ts
@@ -1,0 +1,24 @@
+export function fileDropFilter(
+  fileList: FileList,
+  acceptedTypes: string[] | undefined,
+  multiple: boolean,
+) {
+  const validFiles = Array.from(fileList).filter(file => isFileTypeAccepted(file, acceptedTypes));
+
+  const limitedFiles = multiple ? validFiles : validFiles.slice(0, 1);
+
+  return limitedFiles.length > 0 ? filesToFileList(limitedFiles) : null;
+}
+
+function isFileTypeAccepted(file: File, acceptedTypes: string[] | undefined) {
+  // allow all file types if no types are specified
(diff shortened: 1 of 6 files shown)


marcjulian:
@ashley-hunter please have a look if this looks like you want to include for the `NgpFileUpload` and also `NgpFileDropzone`. I only added it to `NgpFileUpload` to get your feedback first. How to you like to share the logic between `NgpFileUpload` and `NgpFileDropzone`?

ashley-hunter:
> @ashley-hunter please have a look if this looks like you want to include for the `NgpFileUpload` and also `NgpFileDropzone`. I only added it to `NgpFileUpload` to get your feedback first. How to you like to share the logic between `NgpFileUpload` and `NgpFileDropzone`?

Hey! Thank you so much for the contribution! I completely forgot about that case, but of course you are correct that should be there!

Yes, I think it would be great to have this in both, perhaps we could just create a file (probably in the drop zone folder) that contains some utility functions that could be shared between both? I guess you would need to pass the files, multiple state and the fileTypes to them as parameters rather than referring to "this", but I'm fine with that. Does that sound reasonable to you?

packages/ng-primitives/file-upload/src/file-upload/file-upload.ts
```diff
@@ -190,8 +190,32 @@ export class NgpFileUpload {
...
+    const fileList = event.dataTransfer?.files;
+    if (fileList) {
+      const validFiles = Array.from(fileList).filter(file => this.isFileTypeAccepted(file));
+
+      const limitedFiles = this.state.multiple() ? validFiles : validFiles.slice(0, 1);
+
+      if (limitedFiles.length > 0) {
+        this.selected.emit(this.filesToFileList(limitedFiles));
+      } else {
+        this.selected.emit(null);
```
ashley-hunter: Should we emit at all in this case? Maybe there is a reason why we should, just curious to hear your thoughts 😊

marcjulian: Yes i am also not sure if this emit is necessary. Idea was that when the users drops 3 images but only pdfs are allowed, nothing well happen because they are filtered out. 
We could provide an output which can be used to inform the user with alert/toast etc. that files have been filtered out or are not allowed.

marcjulian:
I extracted it into `fileDropFilter` to use it in dropzone and upload.

packages/ng-primitives/file-upload/src/file-upload/file-upload.ts
ashley-hunter: Ah yes, that would indeed be a good use case! We have an ngpFileUploadCanceled event, do you think that would be a good candidate? Or should we add a separate ngpFileUploadRejected or something similar? (And same for drop zone)

marcjulian: I like the separation with `ngpFileUploadRejected`, because it differs from the canceled event and thus the message may vary too.

ashley-hunter: Perfect! I tend to agree! Let's go with that! Thank you!

marcjulian: I added the outputs and updated the examples.

packages/ng-primitives/file-upload/src/file-dropzone/file-drop-filter.ts
```diff
@@ -0,0 +1,24 @@
...
+  const limitedFiles = multiple ? validFiles : validFiles.slice(0, 1);
+
+  return limitedFiles.length > 0 ? filesToFileList(limitedFiles) : null;
+}
+
+function isFileTypeAccepted(file: File, acceptedTypes: string[] | undefined) {
+  // allow all file types if no types are specified
+  if (!acceptedTypes || acceptedTypes.length === 0) return true;
+
+  return acceptedTypes.some(acceptedType => file.type.match(acceptedType));
```
marcjulian: I'll need to improve the matching here better. Currently doesn't match `.jpg` correctly

ashley-hunter (approved):
Awesome! Great work as always! Thank you so much for the contribution!

ashley-hunter:
@allcontributors add @marcjulian for code and ideas

PR #253, opened 2025-05-20
feat: add option to disable escape key

### LINKED ISSUE #216: dialog: option to disable escape key
### Feature Type

Enhancement

### Feature Description

We should allow the escape key to be disabled

kedevked:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

Closes #216

- [ ] Yes
- [x] No

### DIFF
--- a/packages/ng-primitives/dialog/src/dialog-overlay/dialog-overlay.ts
+++ b/packages/ng-primitives/dialog/src/dialog-overlay/dialog-overlay.ts
@@ -1,5 +1,5 @@
 import { BooleanInput } from '@angular/cdk/coercion';
-import { booleanAttribute, Directive, HostListener, input } from '@angular/core';
+import { booleanAttribute, Directive, effect, HostListener, input } from '@angular/core';
 import { NgpExitAnimation } from 'ng-primitives/internal';
 import { injectDialogRef } from '../dialog/dialog-ref';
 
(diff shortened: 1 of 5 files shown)


packages/ng-primitives/dialog/src/dialog-overlay/dialog-overlay.ts
```diff
@@ -21,6 +21,23 @@ export class NgpDialogOverlay {
...
+    transform: booleanAttribute,
+  });
+
+  disableEscapeKeyEffect = effect(() => {
+    if (this.disableEscapeKey()) {
+      this.dialogRef.disableEscapeKey = true;
+    } else {
+      this.dialogRef.disableEscapeKey = false;
+    }
+  });
```
ashley-hunter: I had some time to look into this further - and I think it would be better if this was moved to the dialog-trigger component - I told you to put it here, so that was bad advice on my part. It makes sense for click as the click is actually occurring on the backdrop, but the escape key listening is not an overlay thing. If we move it to the trigger we can simply pass it as config to the `open` method, which will mean we don't need any effects or things like that.

I hope that makes sense.

ashley-hunter (changes requested):
Thank you! Change looks great, I just had a rethink about the best place to put this, but apart from that looks great.

packages/ng-primitives/dialog/src/dialog-overlay/dialog-overlay.ts
kedevked: It definitely makes sense not to be in  `dialog-overlay`. However I wonder if `disableClose` and `disableEscapeKey` carry the same weight more or less. if so why disableClose is not in `dialog-trigger` ?

ashley-hunter (approved):
Great! Thank you!

PR #271, opened 2025-05-30
docs: toggle example style

### LINKED ISSUE #146: add option to toggle example styles in docs code snippets
### Describe the problem that you experienced

Currently all our examples contain styles that rely on our example theme for variables. This is often helpful, but we should allow the styles to be hidden. We should retain the selectors, just remove the rules inside the blocks.

This should apply to both snippets and reusable components snippets.

kedevked:
## PR Checklist

- [ ] The commit message follows our guidelines: https://github.com/ng-primitives/ng-primitives/blob/main/CONTRIBUTING.md#-commit-message-guidelines

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

Closes #146 

- [ ] Yes
- [x] No


kedevked:
When a component has several styles, they are display as follows:

image

For components with only one style, the select is not shown

ashley-hunter:
Hey! Thank you! Nice work! Looks great! And at a quick glance the setup you have in place looks great - especially for potential future tailwind support to!

I wonder for unstyled variants might we be able to remove the styles automatically - rather than needing to explicitly add examples for them all. For tailwind this automated approach would not be possible and your setup is perfect!

We have code to strip styles out in our generator (https://github.com/ng-primitives/ng-primitives/blob/next/packages/ng-primitives/schematics/ng-generate/index.ts#L47), I'm wondering could we do something similar here? I'm happy to help if needed.

But, nice work, this will set us up nicely for the future! Thank you!

kedevked:
```ts
code.replace(/styles\s*:\s*`[^`]*`,?\s*/, '')
```
It will do the trick for what is displayed. However, there is also the `component` that is displayed.

```html
 <ng-container [ngComponentOutlet]="component()" />
```

I thought about striping the style initially. But since the code is generated, I think that it won't be an issue to have an unstyled component as a separate file

kedevked:
Okay, now I see. I will make the changes later during the weekend

ashley-hunter:
Absolutely! This setup is the tricky bit! Adding the examples are great tasks for anyone who wants to contribute for the first time or get familiar with the project!

ashley-hunter:
Sorry I didn't see them.

The important thing is the source code, so as long as that updates that is the main thing.

As mentioned an unstyled preview will just look broken, so I don't think we should show that.

Ideally the tailwind and css examples should look identical, so switching the previews is not essential - that being said, for developing purposes it would be nice to be able to see the tailwind example in the preview in the docs particularly when developing, so you know that your styles look as expected.

It would be best if your preference was saved in local storage so you don't have to change it each time. So if we consider that, it's probably not a great experience for someone to see "no preview" each time they go to a page if their preference is unstyled.

So ideally I think the best case would be:

CSS - Show standard preview 
Tailwind - Show Tailwind preview
Unstyled - Show standard preview

i hope that makes sense

ashley-hunter (approved):

PR #298, opened 2025-06-19
fix: sync state and data-attributes in form-field and form-control - control status

adriankopytko:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

Closes #

This PR introduces two fixes for form-field and form-control

### `form-field` now correctly syncs data-attributes and its state
Previous solution was updating state based on `valueChanges` stream, which would not update state and data-attribs on events such as `touch`, without changing value of the control, i.e. user clicking input and immediately clicking away. My proposed solutions uses recently introduced `events` stream on internal ngControl's control which reacts to more events, more can be found here: https://github.com/angular/angular/issues/10887.
Take note that:
- `statusChanges` wouldn't work as it primarily emits when control's validity changes, not other events
- `events` adds slight perf overhead to the solution, as it may emit multiple times (i.e. when value and validity change at the same time). It is mostly neglegible, as most of the signals are updated with primitive values and won't be updated anyway, except `errors` which is an array

### `form-control` now correctly syncs data-attributes and its state, including its initialization and ngControl.control recognition
Setting up form control had three main issues, which made it more challenging that form-field:
- initialization of state and data-attribs didn't work sometimes (there were no data-attribs at all)
- some examples in the docs were working correctly, with correct initial state and partially correct updates, and some were not
- even if they were updating, updates only happened on `statusChange` (so mainly form-field case)

I've spent some time and figured out that there is a fundamental difference in how you should handle controls applied via `ngModel` and `formControl|formControlName` directives. Currently, i.e for input directive, we are setting up form control in `constructor`, which works fine with `ngModel` as it creates its internal formcontrol right away. Unfortunately, `formControl|formControlName` directives set up their controls in `ngOnChanges`, which is updated roughly at the same time as `ngOnInit` for these directives, but still later than our `constructor` form control setup. That caused it to never subscribe to changes or set its init state correctly for controls that used reactive forms (ngControl was never fully initialized at this point in time, so we didn't have access to its state, streams or underlying control). I'm not sure if my proposed solutions is the best available, but I decided to implement it with the use of `afterRenderEffects` as we are quite sure that at this point all controls that can be reached are available. This way we can successfully set initial state and subscribe in reactive forms.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/ng-primitives/utils/src/forms/status.ts
+++ b/packages/ng-primitives/utils/src/forms/status.ts
@@ -1,4 +1,4 @@
-import { Signal, inject, signal } from '@angular/core';
+import { DestroyRef, Signal, WritableSignal, afterNextRender, inject, signal } from '@angular/core';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { NgControl } from '@angular/forms';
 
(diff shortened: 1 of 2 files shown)


packages/ng-primitives/utils/src/forms/status.ts
```diff
@@ -32,20 +73,22 @@ export function controlStatus(): Signal<NgpControlStatus> {
 
   // Fallback if control is not yet available
   if (!control?.control) {
+    // There is still a chance that the control will be available i.e. after executing OnInit lifecycle hook
+    // in `formControlName` directive, so we set up an effect to subscribe to the control status
+    afterRenderEffect({
```
ashley-hunter: Would it be more appropriate here to use afterNextRender as we are not looking to call this repeatedly any time a signal changes from what I can tell. I think in this case, we're just looking to run this after the initial render has completed.

ashley-hunter (changes requested):
Thank you very much for the contribution, great work! I just had one query about how waiting for rendering is handled? Apart from that looks great.

packages/ng-primitives/utils/src/forms/status.ts
adriankopytko: Yes, you are right, for some reason I thought `afterRenderEffect` would do the same at 4am 😄

PR #343, opened 2025-07-07
feat(popover): add opened and closed event outputs to NgpPopoverTrigger

eurusik:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

Closes #

This PR adds event outputs to the NgpPopoverTrigger directive:

- Added `opened` output event that emits when the popover is opened
- Added `closed` output event that emits when the popover is closed
- Updated documentation with examples of how to use these events
- Added tests to verify the events are emitted correctly

These events complement the existing data attributes approach by providing a programmatic way to react to popover state changes, allowing developers to perform actions when popovers open or close.

- [ ] Yes
- [x] No


packages/ng-primitives/popover/src/popover-trigger/popover-trigger.ts
```diff
@@ -172,6 +173,16 @@ export class NgpPopoverTrigger<T = null> implements OnDestroy {
...
+  /**
+   * Event emitted when the popover is opened.
+   */
+  readonly opened = output<void>();
+
+  /**
+   * Event emitted when the popover is closed.
+   */
+  readonly closed = output<void>();
+
```
ashley-hunter: Could we simplify this so it is just one output that emits a boolean state. This would make it easy to bind the open state to a variable for example: `(ngpPopoverTriggerOpenChange)="isOpen = $event".

We also make all our directive inputs and outputs start with the directive selector as this helps avoid collisions, so `ngpPopoverTriggerOpenChange` would be the name.

```suggestion
  /**
   * Event emitted when the popover open state changes
   */
  readonly openChange = output<boolean>({
      alias: 'ngpPopoverTriggerOpenChange'
  });

```

packages/ng-primitives/popover/src/popover-trigger/popover-trigger.ts
```diff
@@ -213,7 +224,13 @@ export class NgpPopoverTrigger<T = null> implements OnDestroy {
...
 
     // Show the overlay
-    this.overlay()?.show();
+    this.overlay()
+      ?.show()
+      .then(() => {
+        if (this.open()) {
+          this.opened.emit();
+        }
+      });
```
ashley-hunter: I think we should switch this to use async/await - it would be better for the consumer that they can await the method to know when it has actually opened - I should have done this before 😊

```suggestion
    await this.overlay()?.show();
    this.openChange.emit(true);
```

ashley-hunter (changes requested):
Hey, thank you very much! Just had a few small suggestions. Nice work!

packages/ng-primitives/popover/src/popover-trigger/popover-trigger.ts
eurusik: Thank you for the feedback! I've implemented both suggested changes:

1. Simplified the event API:
   - Replaced separate `opened` and `closed` events with a single `ngpPopoverTriggerOpenChange` event that emits a boolean value
   - This makes it easier for consumers to bind to state variables with [(ngpPopoverTriggerOpenChange)="isOpen = $event"](cci:1://file:///c:/dev/ng-primitives/packages/ng-primitives/popover/src/popover-trigger/popover-trigger.ts:231:2-246:3)
   - Updated documentation and tests to reflect this change

2. Improved async handling:
   - Converted both [show()](cci:1://file:///c:/dev/ng-primitives/packages/ng-primitives/popover/src/popover-trigger/popover-trigger.ts:208:2-229:3) and [hide()](cci:1://file:///c:/dev/ng-primitives/packages/ng-primitives/popover/src/popover-trigger/popover-trigger.ts:231:2-246:3) methods to use async/await pattern
   - Added return type `Promise<void>` to both methods
   - Updated JSDoc to indicate that the methods return promises
   - This allows consumers to await these methods to know when operations are complete

All tests are passing. Let me know if you'd like any further adjustments!

ashley-hunter (approved):
All good, thank you!

mauricio-camacho:
Hi there! This is really nice! Just a small suggestion: I think it would be also nice if we emitted the change event whenever the popover is closed, e.g., when clicking outside or when the escape key is pressed. As it is now, we're only emitting the event when we click the popover trigger.

mauricio-camacho:
Will do @ashley-hunter! Thanks for the quick response and keep up with the great job on this library!
I have one last related question though, is there currently a way we can programmatically close a popover?

PR #363, opened 2025-07-14
fix(form-control): apply native disabled attribute to ngpFormControl host element

### LINKED ISSUE #362: ngpFormControl disabling not working
### Affected Primitive

Form-Field

### Version

0.63.0

### Bug Description

The ngpFormControl directive does not properly apply the disabled state to the underlying form control element. Despite setting the ngpFormControlDisabled input or the control being marked as disabled via parent state, the native disabled attribute is not reflected on the element, resulting in a control that appears enabled and is still focusable/interactable.

### Steps to Reproduce

1.	Use <input ngpFormControl [ngpFormControlDisabled]="true" /> in a template.
2.	Observe that the input is still interactive and does not have the native disabled behavior.	

### Reproduction

_No response_

kostetskyroma:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

Closes #362 

This PR fixes an issue where the ngpFormControl directive was not properly disabling native form elements.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/ng-primitives/form-field/src/form-control/form-control.ts
+++ b/packages/ng-primitives/form-field/src/form-control/form-control.ts
@@ -4,6 +4,8 @@ import {
   booleanAttribute,
   computed,
   Directive,
+  HOST_TAG_NAME,
+  inject,
   input,
   signal,
   Signal,
@@ -12,6 +14,12 @@ import { explicitEffect, injectElementRef } from 'ng-primitives/internal';
 import { controlStatus, NgpControlStatus, uniqueId } from 'ng-primitives/utils';
 import { injectFormFieldState } from '../form-field/form-field-state';
 import { formControlState, provideFormControlState } from './form-control-state';
+
+/**
+ * Custom elements that support the disabled attribute.
+ * Example: 'mat-select', 'ng-select'
+ */
+const CUSTOM_ELEMENTS_WITH_DISABLED_ATTRIBUTE: string[] = [];
 
 /**
  * Typically this primitive would be not be used directly, but instead a more specific form control primitive would be used (e.g. `ngpInput`). All of our form control primitives use `ngpFormControl` internally so they will have the same accessibility features as described below.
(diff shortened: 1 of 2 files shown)


packages/ng-primitives/form-field/src/form-control/form-control.ts
```diff
@@ -22,6 +22,9 @@ import { formControlState, provideFormControlState } from './form-control-state'
   selector: '[ngpFormControl]',
   exportAs: 'ngpFormControl',
   providers: [provideFormControlState()],
+  host: {
+    '[attr.disabled]': 'status().disabled ? "" : null',
+  },
```
ashley-hunter: I think we need to make one additional check here, the disabled input is only valid on certain HTML elements, we should check to make sure we only apply it where valid.

ashley-hunter (changes requested):
Thank you very much for the contribution! Looks good, just noticed one additional check we should make

packages/ng-primitives/form-field/src/form-control/form-control.ts
kostetskyroma: This makes me consider changing the selector property to apply ngpFormControl only where it is allowed:

```
selector: 'input[ngpFormControl], select[ngpFormControl], textarea[ngpFormControl], button[ngpFormControl], optgroup[ngpFormControl], option[ngpFormControl]',
```

This way, we filter out all ineligible uses of the ngpFormControl directive upfront. As a result, we no longer need to manually check properties like disabled within the directive.

Additionally, if we introduce similar properties in the future, we can avoid adding separate checks for each one.

@ashley-hunter what do you think about it?

ashley-hunter: Thanks! The primary function of the directive is to apply data attributes representing the state to an attribute, and this may or may not be related to an underlying html element or not, so I don't think we want to limit it to just those. For example, in Angular material the form classes and aria attributes get applied to a mat-select element:

image

I think the most flexible way would be to inject the `HOST_TAG_NAME` and use it to detect whether the element supports the attribute and conditionally apply it.

kostetskyroma: @ashley-hunter Thanks for the feedback! Could you check this commit [refactor(form-control): enhance disabled attribute handling for ngpFo…](https://github.com/ng-primitives/ng-primitives/pull/363/commits/e31ec53c06f36583bd589a9fd61b52e683e849be)? Is this what you expected?

ashley-hunter (approved):
Great! Thank you!

PR #398, opened 2025-07-31
feat: allows responsive toast width

eurusik:
# feat(toast): Make width optional to support responsive styling without !important overrides"

This change provides more flexibility in styling toasts, allowing developers to leverage CSS for responsive designs without needing to use `!important` overrides.

## Changes

- Made the `width` property optional in the NgpToastConfig interface
- Updated the toast manager to conditionally apply width styles only when width is specified
- Updated the toast component to conditionally bind CSS variables
- Updated documentation to reflect that width is now optional
- Removed the default fixed width from the default toast config

These changes allow toast width to be controlled via CSS when no width is specified in the configuration, making it easier to create responsive toast notifications that adapt to different viewport sizes.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

Closes #

This PR makes toast width responsive by default when no width is specified in the configuration. Previously, toast width was fixed at 360px by default, which caused issues with centered toast placement and made it difficult to create responsive designs without using `!important` CSS overrides.

Now, developers can:
- Use responsive toast width by omitting the `width` property in configuration
- Continue using fixed width toasts by specifying a width value
- Style toast width using CSS without needing `!important` overrides
- Create better responsive designs that adapt to different viewport sizes

- [ ] Yes
- [x] No

The change is backward compatible as existing code that specifies a width will continue to work as before. Only the default behavior has changed to be responsive when width is omitted.

### DIFF
--- a/apps/documentation/src/app/pages/(documentation)/primitives/toast.md
+++ b/apps/documentation/src/app/pages/(documentation)/primitives/toast.md
@@ -142,7 +142,7 @@ bootstrapApplication(AppComponent, {
   The duration in milliseconds that the toast will be visible.
 </prop-details>
 
-<prop-details name="width" type="number | undefined" default="undefined">
+<prop-details name="width" type="number | undefined" default="360">
   The width of each toast in pixels. If not provided, the toast will be responsive based on viewport width, allowing for custom styling via CSS.
 </prop-details>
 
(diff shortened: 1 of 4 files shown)


ashley-hunter (commented):
Hey! Thank you very much for the contribution - thinking about this I wonder how much benefit this width config option is at all, I think it might just be better to remove it all together and just use CSS always. Thoughts?

Ruudt:
If removed that would probably be fine. It is a style choice. The centered placement styles might need updating to remain working then?

ashley-hunter:
Yeah, thats what I think, its a style thing, not really something we need to control programmatically, and it gives much more freedom in terms of responsive beheviour too. 

That is a good, point though, something that would need verified! Thanks!

Issue #82, opened 2024-11-01
Accordian: use `hidden-until-found` internally to support browser native "Find in Page"


ajitzero:
### Feature Type

Enhancement

### Feature Description

- Reference: https://developer.chrome.com/docs/css-ui/hidden-until-found
- Caveat: This is for collapsible sections that exist in the DOM, so not applicable for sections hidden with `ngIf/@if`

**Current Behavior:**

Try searching for the term "open source" on this page: https://angularprimitives.com/primitives/accordion.

**Expected Behavior:**

Browser-native search should work for hidden/collapsed sections.

ashley-hunter:
Thank you very much for the suggestion! I actually came across this same article a week or two ago myself. I haven't implemented it yet because the feature is still marked as `experimental` and has limited browser support:

![image](https://github.com/user-attachments/assets/80802b9d-5183-40fe-bf8f-82f3f2e9f268)
https://caniuse.com/?search=hidden%3Duntil-found%20

But I will leave this issue open as this is definitely something I would like to do as soon as possible. Great suggestion, thank you!

ajitzero:
Yeah, that makes sense! I saw that it has been implemented in React Aria as a part of the Disclosure component so I thought it would be great to include here as well. Plus, it is a progressive enhancement and doesn't impact browsers that don't support it.

ashley-hunter:
Ah interesting, if that's the case then we could probably add this for browsers that support it! Thank you 😊

PR #402, opened 2025-08-01
feat(accordion): add hidden until found support in accordion content

RaimonxDev:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

Closes #
#82 

The `ngpAccordionContent` primitive uses the `until-found` attribute to allow the browser to search text within the hidden region and reveal the section if a match is found. If the browser does not support this functionality, the attribute is ignored

- [ ] Yes
- [x] No


packages/ng-primitives/accordion/src/accordion-content/accordion-content.ts
```diff
@@ -56,6 +63,34 @@ export class NgpAccordionContent<T> {
...
+     * The hidden attribute must be removed.
+     */
+    afterRenderEffect({
+      write: () => {
+        if (!this.supportsBeforeMatch()) return;
+
+        if (this.accordionItem().open()) {
+          this.elementRef.nativeElement.removeAttribute('hidden');
+        } else {
+          this.elementRef.nativeElement.setAttribute('hidden', 'until-found');
```
ashley-hunter: I notice that whenever we collapse the accordion (if it's animated) the content disappears immediately - I wonder is there any way, to only have the content hidden whenever the height is 0, perhaps a resize observer could be useful for this?

Let me know if this makes sense.

ashley-hunter (changes requested):
Thank you very much for the contribution! I just noticed one issue where the content can disappear immediately when there is an animation.

packages/ng-primitives/accordion/src/accordion-content/accordion-content.ts
RaimonxDev: I have solved the problem of observing the height with “fromResizeEvent.” I look forward to your comments.

packages/ng-primitives/accordion/src/accordion-content/accordion-content.ts
```diff
@@ -42,6 +44,11 @@ export class NgpAccordionContent<T> {
    */
   readonly id = input<string>(uniqueId('ngp-accordion-content'));
 
+  /**
+   * The platform id
+   */
+  private readonly platformId = inject(PLATFORM_ID);
```
ashley-hunter: Thank you, changes look great! We have just merged updates to a utility which I think would simplify this. We can use the `injectDimensions` utility to get the height as a signal, which means the hidden attribute value could be a computed. I believe this would also remove the need to check if it is supported in the browser too and the manual dom updates. Something like this I believe:

```ts
import { afterRenderEffect, computed, Directive, input } from '@angular/core';
import { fromMutationObserver, injectDimensions, injectElementRef } from 'ng-primitives/internal';
import { safeTakeUntilDestroyed, uniqueId } from 'ng-primitives/utils';
import { debounceTime } from 'rxjs';
import { injectAccordionItemState } from '../accordion-item/accordion-item-state';
import type { NgpAccordion } from '../accordion/accordion';
import { injectAccordionState } from '../accordion/accordion-state';

/**
 * Apply the `ngpAccordionContent` directive to an element that represents the content of an accordion item.
 */
@Directive({
  selector: '[ngpAccordionContent]',
  exportAs: 'ngpAccordionContent',
  host: {
    role: 'region',
    '[id]': 'id()',
    '[attr.data-orientation]': 'accordion().orientation()',
    '[attr.data-open]': 'accordionItem().open() ? "" : null',
    '[attr.data-closed]': 'accordionItem().open() ? null : ""',
    '[attr.aria-labelledby]': 'accordionItem().triggerId()',
    '[attr.hidden]': 'hidden()',
    '(beforematch)': 'onBeforeMatch()',
  },
})
export class NgpAccordionContent<T> {
  /**
   * Access the accordion content element reference
   */
  private readonly elementRef = injectElementRef();

  /**
   * Access the accordion
   */
  protected readonly accordion = injectAccordionState<NgpAccordion<T>>();

  /**
   * Access the accordion item
   */
  protected readonly accordionItem = injectAccordionItemState();

  /**
   * The id of the content region
   */
  readonly id = input<string>(uniqueId('ngp-accordion-content'));

  /**
   * The dimensions of the content region
   */
  private readonly dimensions = injectDimensions();

  /**
   * The hidden attribute is used for compatibility with the beforematch event
   * and is set to 'until-found' when the content is not open and has no height.
   * This allows the browser's find-in-page functionality to reveal the content when it matches
   * a search term, even if the content is not currently visible.
   */
  protected readonly hidden = computed(() =>
    !this.accordionItem().open() && this.dimensions().height === 0 ? 'until-found' : null,
  );

  constructor() {
    this.accordionItem().content.set(this);

    // any time the open state of the accordion item changes, update the dimensions
    afterRenderEffect(() => this.updateDimensions());

    // update dimensions when the content changes
    fromMutationObserver(this.elementRef.nativeElement, {
      childList: true,
      subtree: true,
      disabled: computed(() => !this.accordionItem().open()),
    })
      .pipe(debounceTime(0), safeTakeUntilDestroyed())
      .subscribe(() => this.updateDimensions());
  }

  /**
   * Handle the beforematch event to automatically open the accordion item
   * when the browser's find-in-page functionality tries to reveal hidden content.
   */
  protected onBeforeMatch(): void {
    const isDisabled = this.accordion().disabled() || this.accordionItem().disabled();

    if (isDisabled) {
      return;
    }
    this.accordion().toggle(this.accordionItem().value() as T);
  }

  private updateDimensions(): void {
    if (this.accordionItem().open()) {
      // remove the inline styles to reset them
      this.elementRef.nativeElement.style.removeProperty('--ngp-accordion-content-width');
      this.elementRef.nativeElement.style.removeProperty('--ngp-accordion-content-height');
      // set the dimensions based on the content
      this.elementRef.nativeElement.style.setProperty(
        '--ngp-accordion-content-width',
        `${this.elementRef.nativeElement.scrollWidth}px`,
      );
      this.elementRef.nativeElement.style.setProperty(
        '--ngp-accordion-content-height',
        `${this.elementRef.nativeElement.scrollHeight}px`,
      );
    }
  }
}
```

Do you think this covers everything?

ashley-hunter: Oh - you would need to update the latest changes to get this function with the latest updates

RaimonxDev: This approach you mentioned is one I had already implemented. Originally, I went with:

```ts
'[attr.hidden]': 'hidden()'
```
It worked, although it had the height bug you mentioned. Other than that, I tested it in Safari, which doesn’t support until hidden, and since the hidden attribute remains on the host, it shows strange behavior in Safari. That’s why I chose to set or remove the property in the afterNextRenderer.

Regarding injectDimensions(), I had already tried it to fix the zero-height issue, but for some reason I don’t understand, it wasn’t reacting to the accordion’s height changes.

With this approach, do you have any performance concerns?

ashley-hunter: Yes, you're absolutely right, this approach would not have worked before, and it would have required the approach you took, it could only work now because of changes I merged last night which is why I just wanted to check if it can be simplified now.

There shouldn't be any performance issues, as the computed is only reading signals and returning a string, so there is no attribute changing a lot that could cause reflow and we aren't manually reading sizes from the dom, the resize observer is taking care of that.

Ah yes, I knew there was some reason for it, I just couldn't recall. If you are still able to see the Safari issue with this approach, then we can keep the supported method, and call it in the computed. I think that should work if it's still needed.

Thank you

ashley-hunter (approved):
Great! Nice work, thank you!

PR #403, opened 2025-08-03
  feat(tooltip): add useTextContent feature for automatic text content tooltips

shaharkazaz:
feat(tooltip): add useTextContent feature for automatic text content tooltips

  - Add useTextContent input to NgpTooltipTrigger (defaults to true)
  - Allow passing strings directly to ngpTooltipTrigger
  - Update documentation with examples and global styles requirement
  - Add test coverage

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Passing a string to a tooltip is a common use case, especially when displaying ellipses for truncated items but still wanting users to see the full content. Currently, this requires extra boilerplate—wrapping the string in a template reference just to display a simple string value.

## Issue

Closes #

With this update, NgpTooltipTrigger now accepts strings as inputs, in addition to `TemplateRef`s and components.

This PR also introduces the `useTextContent` option, which automatically extracts the trigger’s text content and uses it as the tooltip’s value. Since this is a frequent pattern, the new shorthand reduces boilerplate and results in much cleaner code.

It depends on whether you consider setting the useTextContent to true by default breaking or not. Since passing values to the trigger takes precedence, it shouldn't affect users.
The bigger question is rather we want to set it to true by default or not. @ashley-hunter your call.

- [ ] Yes
- [ ] No
- [x] Maybe

### DIFF
--- a/packages/ng-primitives/tooltip/src/tooltip-text-content/tooltip-text-content.component.ts
+++ b/packages/ng-primitives/tooltip/src/tooltip-text-content/tooltip-text-content.component.ts
@@ -1,0 +1,22 @@
+import { Component } from '@angular/core';
+import { injectOverlayContext } from 'ng-primitives/portal';
+import { NgpTooltip } from '../tooltip/tooltip';
+
+
+/**
+ * Internal component for wrapping string content in tooltip portals
+ * @internal
+ */
+@Component({
+  template: '{{ content() }}',
+  hostDirectives: [NgpTooltip],
+  host: {
+    ngpTooltip: '',
+  },
+})
+export class NgpTooltipTextContentComponent {
+  /**
+   * The string content to display
+   */
+  readonly content = injectOverlayContext();
+}
--- a/packages/ng-primitives/tooltip/src/tooltip-trigger/tooltip-trigger.ts
+++ b/packages/ng-primitives/tooltip/src/tooltip-trigger/tooltip-trigger.ts
@@ -22,7 +22,10 @@ import {
   NgpOverlayContent,
 } from 'ng-primitives/portal';
 import { injectTooltipConfig } from '../config/tooltip-config';
+import { NgpTooltipTextContentComponent } from '../tooltip-text-content/tooltip-text-content.component';
 import { provideTooltipTriggerState, tooltipTriggerState } from './tooltip-trigger-state';
+
+type TooltipInput<T> = NgpOverlayContent<T> | string | null | undefined;
 
 /**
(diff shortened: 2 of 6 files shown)


packages/ng-primitives/tooltip/src/tooltip-text-content/tooltip-text-content.component.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Internal component for wrapping string content in tooltip portals
+ * @internal
+ */
+@Component({
+  template: '{{ content() }}',
+  hostDirectives: [NgpTooltip],
+  host: {
+    ngpTooltip: '',
```
shaharkazaz, line 15: I can add a comment here that it's for styling targeting purposes as the host directive isn't actually being added at the DOM and this is our entry point for styling.

packages/ng-primitives/tooltip/src/tooltip-trigger/tooltip-trigger.ts
```diff
@@ -225,18 +239,38 @@ export class NgpTooltipTrigger<T = null> implements OnDestroy {
...
+      const textContent = this.trigger.nativeElement.textContent?.trim() || '';
+      if (ngDevMode && !textContent) {
+        console.warn(
+          '[ngpTooltipTrigger]: useTextContent is enabled but trigger element has no text content',
+        );
+        return;
+      }
+      content = NgpTooltipTextContentComponent;
+      context = signal(textContent);
+    } else if (typeof content === 'string') {
```
shaharkazaz: @ashley-hunter I wasn't sure if you wanted to introduce a new validators.ts helper for these kinds of assertions, as I saw them inlined in the repo. LMK

ashley-hunter: Yeah, that might be a good idea, there are a few places this kind of check is made so we could reduce duplication.

packages/ng-primitives/tooltip/src/tooltip-text-content/tooltip-text-content.component.ts
ashley-hunter: Makes sense, a comment may be good just so I don't forget the reasoning for it, but good idea!

shaharkazaz:
@ashley-hunter In your free time 🙂

PR #407, opened 2025-08-06
feat(date-picker): add option to change the first day of the week for…

### LINKED ISSUE #405: Allow configuring first day of week (e.g. Monday) in date-picker / date-range-picker
### Feature Type

Enhancement

### Feature Description

Add a firstDayOfWeek option (or an option to auto-detect from the `LOCALE_ID`) so weeks can start on Monday instead of Sunday. When using a Monday-first calendar (e.g. in Germany), the current date (range) picker always starts weeks on Sunday, which is not ideal. 

When adding this, the docs should point out that you'll have to reorder the header columns from the example by hand, otherwise the days won't match obviously (move Sunday next to Saturday).

Thanks so much for this library - really enjoying it! Happy to submit a PR once there’s consensus on the API.

bboehm86:
… the calendar

This adds a new input option for `NgpDatePicker` and `NgpDateRangePicker`and also introduce a `NgpDatePickerConfig` to set a global default configuration option.

This does not change the default behavior.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

Closes #405

This adds the option to change the start of the week for the `NgpDatePicker` and `NgpDateRangePicker` calendar. It's necessary since most international applications use Monday as the start of the week, and not the current default of Sunday.

The first day of the week can be set via the newly introduced `ngpDatePickerFirstDayOfWeek` / `ngpDateRangePickerFirstDayOfWeek` inputs via a number (`0-7`) or by providing a default value with the `provideDatePickerConfig()` function.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/ng-primitives/date-picker/src/date-picker-row-render/date-picker-row-render.ts
+++ b/packages/ng-primitives/date-picker/src/date-picker-row-render/date-picker-row-render.ts
@@ -161,6 +161,8 @@ export class NgpDatePickerRowRender<T> implements OnDestroy {
    * Get the offset of the first day of the week.
    * @param firstCalendarDay The first day of the calendar without the offset.
    * @returns The offset of the first day of the week.
+   *
+   * @internal
    */
   getFirstDayOfWeekOffset(firstCalendarDay: T): number {
     return (


packages/ng-primitives/date-picker/src/date-picker-row-render/date-picker-row-render.ts
```diff
@@ -92,11 +103,16 @@ export class NgpDatePickerRowRender<T> implements OnDestroy {
   private readonly viewRefs: EmbeddedViewRef<void>[] = [];
 
   constructor() {
-    // re-render the rows when the month changes.
-    onChange(this.state().focusedDate, (date, previousDate) => {
-      if (!date || !previousDate || !this.dateAdapter.isSameMonth(date, previousDate)) {
-        this.renderRows();
-      }
+    // Wait for the inputs of the containing picker to be initialized.
+    afterNextRender(() => {
```
bboehm86: I assumed here that the value of the `firstDayOfWeek` input won't change after the first initialization and that we do not need to handle those cases. But we have to wait for the population of the inputs, so I moved this into the `afterNextRender()` here. I'm open for any better ideas

ashley-hunter: Thanks, I pushed one change to simplify the logic that was already there, after that your changes seem to work perfectly 😊

ashley-hunter:
@allcontributors add @bboehm86 for code

packages/ng-primitives/date-picker/src/date-picker-row-render/date-picker-row-render.ts
bboehm86: Awesome, thanks for the quick review and also release! 🙂

ashley-hunter: No problem! Thanks for the contribution 😊

PR #409, opened 2025-08-06
feat(menu): add right-click support to menu trigger

liranmizrahi1:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

Adds the ability to trigger the menu with a right-click. A new `ngpMenuTriggerOn` input has been added to control this behavior.

The possible values are:
- `left` (default): triggers the menu on a left-click.
- `right`: triggers the menu on a right-click.
- `both`: triggers the menu on both left and right-clicks.

When `right` or `both` is used, the default browser context menu is prevented from opening.

- [ ] Yes
- [x] No


packages/ng-primitives/menu/src/menu-trigger/menu-trigger.ts
```diff
@@ -23,6 +23,7 @@ import {
 import { injectMenuConfig } from '../config/menu-config';
```
ashley-hunter, line 23: I think this might be good to split out into a separate directive - perhaps a context-menu-trigger. 

It would largely be the same code, and I'm ok for duplicating most of it but I think there are some additional differences to consider, not just the launch interaction - as context menus usually launch relative to the mouse position rather than a trigger element e.g.:

https://ui.shadcn.com/docs/components/context-menu

Having this as a separate primitive would allow us to properly handle these differences without having to add lots of conditions in the menu primitive.

Any thoughts?

liranmizrahi1: Hi, you made a really good point 👍.
Actually, my intention was to have the item itself act as the menu trigger — for example, a text element inside a grid that uses the grid’s row data as its context.
That way, the text would serve as the trigger element by right click.

I think we could add the input to the menu trigger and create a new context-menu-trigger directive, similar to this example:
https://ui.shadcn.com/docs/components/context-menu

ashley-hunter: Thank you! I’ve been looking into how other headless UI libraries handle this scenario, and interestingly, I couldn’t find any that build this functionality directly into their menu components. Instead, they all tend to keep this separate in a context menu primitive. This seems to be a cleaner separation of concerns and probably the more maintainable approach in the long run.

That said, your use case does highlight a valid limitation in our current MenuTrigger design. Rather than adding built-in right-click support, I think we could address this more broadly by allowing programmatic control of the menu.

For example, if `ngpMenuTriggerDisabled` disabled only the default event handlers (click, enter, etc.) but still allowed calling `show()` manually, it would open up more flexible use cases - not just right-click, but also double-click, keyboard shortcuts, or any other interaction. This way, we avoid adding one-off inputs for every possible trigger, but still make the component highly customizable.

liranmizrahi1: Your idea is excellent! 
but I need to see how this would work for my specific case though
 I have one menu with its reference passed to multiple triggers

ashley-hunter: I think it would simply be something like this:

https://stackblitz.com/edit/wszxsj76?file=src%2Fmain.ts

The change in disabled state would not be required after we make the change to the trigger, but should be fairly simple.

liranmizrahi1: Amazing I will close this pr
do you me to take this task(change disable and create context menu directive)?

ashley-hunter: That would be great! Thank you very much!

Issue #277, opened 2025-06-03
Feature Request: Support for Range Slider (Double Thumb) in Slider Primitive


Damian96:
### Feature Type

New Primitive

## Description
It would be extremely useful to have native support for a range slider (i.e., a double-thumb slider) in angular-primitives. This enhancement would allow users to select a value range — such as a price filter — with a visual interface and single component. Could also be a new separate primitive instead of extending the current slider primitive ( I'm not sure what would be the best practice here )

## Use Cases
- E-commerce Filters: Users can select a price range
- Custom Thresholds: Users can define operating ranges or limits in dashboards

## Related Work
- Libraries like @angular-slider/ngx-slider, React’s @radix-ui/react-slider already support this behavior.

Thank you for the amazing work on Angular Primitives!

ashley-hunter:
Thank you! Glad you like the project! Much appreciated! Yes, this is something I would really like to add! I will certainly try to add it soon, thank you!

PR #430, opened 2025-08-16
feat: Range Slider

RaimonxDev:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation changes

Hi Ashley, please review this PR and share your feedback. For now, I’m leaving it as a draft.

I created a range slider inside the slider directory.

I also updated the slider documentation to include the range slider. This is what I have so far:

```html
<div ngpRangeSlider>
  <div ngpRangeSliderTrack>
    <div ngpRangeSliderRange></div>
  </div>
  <div ngpRangeSliderThumb ngpRangeSliderThumbType="low"></div>
  <div ngpRangeSliderThumb ngpRangeSliderThumbType="high"></div>
</div>
```
I look forward to your comments for improvements and corrections. At the moment, everything seems to be working fine.
## Issue
#277 
Closes #

- [ ] Yes
- [ ] No


packages/ng-primitives/slider/src/range-slider/range-slider-thumb/range-slider-thumb.ts
```diff
@@ -0,0 +1,161 @@
...
+  };
+
+  /**
+   * Get the current percentage for this thumb.
+   */
+  protected currentPercentage = () => {
+    return this.thumbType() === 'low'
+      ? this.state().lowPercentage()
+      : this.state().highPercentage();
+  };
```
ashley-hunter: Could we make these computeds?

packages/ng-primitives/slider/src/range-slider/range-slider-thumb/range-slider-thumb.ts
```diff
@@ -0,0 +1,161 @@
...
+  @HostListener('keydown', ['$event'])
+  protected handleKeydown(event: KeyboardEvent): void {
+    const multiplier = event.shiftKey ? 10 : 1;
+    const currentValue = this.currentValue();
+    const step = this.state().step() * multiplier;
+
+    let newValue: number;
+
+    switch (event.key) {
+      case 'ArrowLeft':
```
ashley-hunter, line 145: We may need to also take into account the document direction - RTL for example, ArrowLeft would increase the value rather than decrease.

packages/ng-primitives/slider/src/range-slider/range-slider-thumb/range-slider-thumb.ts
RaimonxDev: Done!

packages/ng-primitives/slider/src/range-slider/range-slider-thumb/range-slider-thumb.ts
RaimonxDev: Hi, I don’t have experience adding RTL support. Could you guide me a bit? From what I understand, do you now need me to invert the arrow keys for increase and decrease? Will anything else change? Should low be on the right and high on the left, or is that not necessary?

packages/ng-primitives/slider/src/range-slider/range-slider/range-slider.ts
```diff
@@ -0,0 +1,184 @@
...
+    const distanceToLow = Math.abs(value - this.state.low());
+    const distanceToHigh = Math.abs(value - this.state.high());
+
+    return distanceToLow <= distanceToHigh ? 'low' : 'high';
+  }
+
+  constructor() {
+    afterNextRender({
+      earlyRead: () => {
+        this.thumbs().forEach((thumb, index) => {
```
RaimonxDev: Regarding the automatic stopping of the thumbs, I implemented the following approach. What do you think? Do you see any alternative solution?

packages/ng-primitives/slider/src/range-slider/range-slider-thumb/range-slider-thumb.ts
ashley-hunter: Hey, thanks for the updates! I have pushed a change which should add support for this, its fairly simple, just need to take into account the direction and swap arrow keys etc. Everything else should be the same as we use inset-inline-start for positioning the thumbs which is a logical property (i.e. it takes document direction into account), so the position should work correctly with no changes!

ashley-hunter (approved):
Looks great! Thank you very much! Nice work 😊 great addition!

PR #494, opened 2025-10-20
fix(select): apply data-selected attribute in multi-select mode

heddendorp:
## Description

Fixes the `data-selected` attribute not being applied to selected options in multi-select mode.

## Problem

The `selected` computed property in `NgpSelectOption` was incorrectly checking if the **option's value** was an array instead of checking if the **select's value** was an array. This caused the selected state to always return `false` in multi-select mode, preventing the `data-selected` and `aria-selected` attributes from being applied.

## Solution

Updated the logic to correctly:
1. Store the select's value (which should be an array in multi-select mode)
2. Check if the select's value is an array
3. Check if this option's value exists in the array of selected values
4. Pass the correct arguments to `compareWith` (option value, then selected value)

## Impact

- ✅ Single-select mode continues to work as before
- ✅ Multi-select mode now correctly applies `data-selected=""` to selected options
- ✅ Multi-select mode now correctly applies `aria-selected="true"` to selected options


packages/ng-primitives/select/src/select-option/select-option.ts
```diff
@@ -69,8 +69,9 @@ export class NgpSelectOption implements OnInit, OnDestroy, NgpActivatable {
     }
 
     if (this.state().multiple()) {
+      const selectValue = this.state().value();
```
ashley-hunter, line 72: Well spotted, I think the value read on line 65 is also incorrect, I think we can change line 65 to `this.state().value()` and then we don't need this second value property.

ashley-hunter (changes requested):
Thank you for the fix! Well spotted, I think we can move the fix to the existing value variable 😊 but looks good, thanks!

ashley-hunter (approved):
Ignore my previous comment, all good!

PR #537, opened 2025-11-25
fix/feat(progress): correct reference between ngpProgress and ngpProgressLabel

### LINKED ISSUE #536: NgpProgressLabel does not assign ID attribute to host
### Affected Primitive

Progress

### Version

0.90.0

### Bug Description

When applying the `ngpProgressLabel` directive to a label, an ID is internally generated, but not assigned to the host component the directive is applied to.

As a result, the component that has the `NgpProgress` directive points to this ID with the `aria-labelledby` attribute, but no node with this ID exists.

### Steps to Reproduce

Example:
```html
<div ngpProgress [ngpProgressValue]="50">
  <span ngpProgressLabel></span>
  <span ngpProgressValue></span>

  <div ngpProgressTrack>
    <div ngpProgressIndicator></div>
  </div>
</div>
```

Resulting HTML:

Image

##### Expected behaviour

The `<span>` with the `ngpProgressLabel` directive has the ID `ngp-progress-label-0`.

##### Actual behaviour

The `<span>` has no ID attribute.

### Reproduction

_No response_

janek64:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

Closes #536 

**Summary:** Implements correct referencing between the `ngpProgress` and `ngpProgressLabel` via HTML attributes that was previously missing.

* Sets the `id` attribute for the host of `ngpProgressLabel` and fixes the `aria-labelledby` attribute in `ngpProgress` to correctly reference it.
* Sets `id` attribute for the host of `ngpProgress` and references it via the `for` attribute of the host of `ngpProgressLabel`.
* Updates the documentation and reusable component to use `<label>` instead of `<span>` for the label so that the `for` attribute can be correctly applied.

- [ ] Yes
- [ ] No

(The reusable component for _Progress_  now uses `<label>` instead of `<span>` - I am not sure if this is considered a breaking change).

### DIFF
--- a/packages/ng-primitives/progress/src/progress-label/progress-label.ts
+++ b/packages/ng-primitives/progress/src/progress-label/progress-label.ts
@@ -1,4 +1,4 @@
-import { Directive, input } from '@angular/core';
+import { Directive, ElementRef, inject, input } from '@angular/core';
 import { uniqueId } from 'ng-primitives/utils';
 import { injectProgressState } from '../progress/progress-state';
 
@@ -7,7 +7,7 @@ import { injectProgressState } from '../progress/progress-state';
   exportAs: 'ngpProgressLabel',
   host: {
     '[attr.id]': 'id()',
-    '[attr.for]': 'state().id()',
+    '[attr.for]': 'elementRef.nativeElement.tagName === "LABEL" ? state().id() : null',
     '[attr.data-progressing]': 'state().progressing() ? "" : null',
     '[attr.data-indeterminate]': 'state().indeterminate() ? "" : null',
     '[attr.data-complete]': 'state().complete() ? "" : null',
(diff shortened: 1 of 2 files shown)


packages/ng-primitives/progress/src/progress-label/progress-label.ts
```diff
@@ -6,6 +6,8 @@ import { injectProgressState } from '../progress/progress-state';
   selector: '[ngpProgressLabel]',
   exportAs: 'ngpProgressLabel',
   host: {
+    '[attr.id]': 'id()',
+    '[attr.for]': 'state().id()',
```
ashley-hunter: Could we possibly add a check here to see if the element is a label element and only add the for attribute when it is? The selector technically allows any element, and people may be using other elements and the for attribute is only valid on labels.

ashley-hunter (approved):
Great work! Thank you very much!

ashley-hunter:
@allcontributors add @janek64 for code.

PR #568, opened 2025-12-12
refactor(progress): state pattern migration

### LINKED ISSUE #557: Progress State Pattern Migration
Migrate progress to the new state patter

zecka:
fix(#557)

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

Closes #557 

- [ ] Yes
- [x] No

### DIFF
--- a/packages/ng-primitives/progress/src/progress/progress-state.ts
+++ b/packages/ng-primitives/progress/src/progress/progress-state.ts
@@ -6,6 +6,11 @@ import { NgpProgressValueTextFn } from './progress';
 
 export interface NgpProgressProps {
   /**
+   * The unique identifier for the progress.
+   */
+  readonly id?: Signal<string>;
+
+  /**
    * Define the progress value.
    */
   readonly value?: Signal<number | null>;
@@ -29,15 +34,31 @@ export interface NgpProgressProps {
    * @returns The value label
    */
   readonly valueLabel?: Signal<NgpProgressValueTextFn>;
+}
 
+export interface NgpProgressState {
   /**
    * The unique identifier for the progress.
    */
-  readonly id?: Signal<string>;
-}
+  readonly id: Signal<string>;
 
-export interface NgpProgressState
-  extends Required<Pick<NgpProgressProps, 'value' | 'min' | 'max' | 'id'>> {
+  /**
+   * Define the progress value.
+   */
+  readonly value: Signal<number | null>;
+
+  /**
+   * Define the progress min value.
+   * @default '0'
+   */
+  readonly min: Signal<number>;
+
+  /**
+   * Define the progress max value.
+   * @default 100
+   */
+  readonly max: Signal<number>;
+
   /**
    * Get the progress value text.
    */


packages/ng-primitives/progress/src/progress/progress-state.ts
```diff
@@ -1,27 +1,139 @@
...
+   */
+  readonly valueLabel?: Signal<NgpProgressValueTextFn>;
+
+  /**
+   * The unique identifier for the progress.
+   */
+  readonly id?: Signal<string>;
+}
+
+export interface NgpProgressState extends Required<NgpProgressProps> {
```
zecka: Not sure about this approach of extending Required<NgpProgressProps>. It might help avoid too much repetition in the property declarations.

packages/ng-primitives/progress/src/progress/progress-state.ts
```diff
@@ -1,27 +1,139 @@
...
+  /**
+   * Get the progress value text.
+   */
+  readonly valueText: Signal<string>;
+
+  /**
+   * The label associated with the progress bar.
+   * @internal
+   */
+  readonly label: Signal<Required<NgpProgressLabelProps> | null>;
```
zecka: Here I keep full props on label like it was before, maybe we can just keep id

ashley-hunter: Yes, just `id` is perfect, thats what I have been doing for other components.

packages/ng-primitives/progress/src/progress/progress-state.ts
ashley-hunter: I think in this case some of the props we don't need to return, valueLabel for example isn't used in any other files, its only used to compute the valueText in this function.

I think we can also remove the following as they are only used in this function and as they are marked as internal in the directive, they can be removed there too.

```ts
  /**
   * Determine if the progress is indeterminate.
   * @internal
   */
  readonly indeterminate = this.state.indeterminate;

  /**
   * Determine if the progress is in a progressing state.
   * @internal
   */
  readonly progressing = this.state.progressing;
  /**
   * Determine if the progress is complete.
   * @internal
   */
  readonly complete = this.state.complete;
```

packages/ng-primitives/progress/src/progress-label/progress-label-state.ts
```diff
@@ -0,0 +1,32 @@
...
+    const element = injectElementRef();
+
+    const state = injectProgressState();
+    state().label.set({ id, ...props });
+
+    attrBinding(element, 'id', id);
+    attrBinding(element, 'for', element.nativeElement.tagName === 'LABEL' ? state().id?.() : null);
+    dataBinding(element, 'data-progressing', () => (state().progressing() ? '' : null));
+    dataBinding(element, 'data-indeterminate', () => (state().indeterminate() ? '' : null));
+    dataBinding(element, 'data-complete', () => (state().complete() ? '' : null));
```
ashley-hunter: The data binding automatically handles booleans for us, so we can remove the ternary:

```suggestion
    dataBinding(element, 'data-progressing', () => state().progressing());
    dataBinding(element, 'data-indeterminate', () => state().indeterminate());
    dataBinding(element, 'data-complete', () => state().complete());
```

packages/ng-primitives/progress/src/progress/progress-state.ts
```diff
@@ -1,27 +1,148 @@
...
+  readonly valueLabel?: Signal<NgpProgressValueTextFn>;
+
+  /**
+   * The unique identifier for the progress.
+   */
+  readonly id?: Signal<string>;
+}
+
+export interface NgpProgressState
+  extends Required<Pick<NgpProgressProps, 'value' | 'min' | 'max' | 'id'>> {
```
zecka: @ashley-hunter 
I’m torn between a KISS approach and a DRY one 😄. Are you ok with this approach, or would you prefer an explicit property declaration even if it means some duplication?
Sorry, this is a lot of small question XD

zecka: I think this image answer my question

ashley-hunter: Yes, I went through the same thought process too 😂 
The conclusion I came to was go with the more verbose approach, yes there is some duplication, but:

1. it makes it super easy to see the full type without jumping about 
2. Progress is kind of the exception, in the sense that most other primitives don't return all props in state. Ideally we only want to return the essential state and no more.
3. It is consistent with all the other primitives.

zecka: Okay perfect, thank's for the details !
The PR is updated !

ashley-hunter (approved):
All looks great! Thank you very much! Great work!

PR #690, opened 2026-03-02
fix(portal): block scroll on ancestor containers and add options to NgpFlipOptions and NgpShiftOptions (#455, #689)

### LINKED ISSUE #455: menu trigger on scrollable element
### Affected Primitive

Menu

### Version

0.71.1

### Bug Description

Issue:
When adding a menu trigger to an element inside a scrollable container, the trigger element can be scrolled out of view while the menu remains open. This results in the menu floating in the wrong place without its trigger visible.

### Expected behavior:
When the menu is open, an overlay should be applied that blocks scrolling on the trigger’s scrollable container, so the trigger cannot disappear while the menu is active.
### Steps to Reproduce

  Place a button with menuTrigger inside a scrollable container.
  
  Open the menu.
  
  Scroll the container so the trigger button moves out of view.
  
  Observe: The menu stays open, detached from the trigger.

### Reproduction

https://stackblitz.com/edit/jyopwtom?file=src%2Fmain.ts

### LINKED ISSUE #689: NgpOverlay flip() uses viewport as boundary - incorrect when trigger is inside a scroll container
### Affected Primitive

NgpOverlay

### Version

0.109.4

### Related

- https://github.com/ng-primitives/ng-primitives/issues/455

### Describe the bug

NgpOverlay runs floating-ui’s `flip()` without a `boundary` option, so it defaults to the viewport (`clippingAncestors`).

When the popover trigger is inside a scrollable container and the popover is portalled to `<body>`, `flip()` sees plenty of viewport space and does not flip, even though the trigger’s scroll container has no room. The popover renders clipped or overflows the scroll container.

### Root cause

In `ng-primitives-portal.mjs`, the middleware pipeline adds a bare `flip()`:

```tsx
if (this.config.flip !== false) {
	middleware.push(flip());
}
```

Because the popover is portalled outside the scroll container into `body`, floating-ui’s default clipping ancestor detection uses `body` and the viewport as the boundary. It does not consider the trigger’s scroll-container ancestors, so it never flips when the scroll container is the constraint.

### Expected behavior

`flip()` should use the trigger element’s overflow ancestors as the boundary so it flips correctly when the trigger is inside a scrollable container.

### Workaround

Disable ng-primitives’ built-in flip and provide custom middleware via `updateConfig`:

```tsx
ngOnInit(): void {
	this.overlay.updateConfig({
		flip: false,
		additionalMiddleware: [
			createBoundaryFlipMiddleware(),
			size({
				apply: ({ availableWidth, availableHeight }) => {
					this.overlay.availableWidth.set(availableWidth);
					this.overlay.availableHeight.set(availableHeight);
				}
			})
		]
	});
}

function createBoundaryFlipMiddleware(): Middleware {
	return {
		name: 'boundaryFlip',
		fn(state: MiddlewareState) {
			const boundary = getOverflowAncestors(state.elements.reference as Element)
				.filter((a): a is Element => a instanceof Element);

			return flip(boundary.length ? { boundary } : undefined).fn(state);
		}
	};
}
```

Note: `size()` is re-run after the custom flip because the built-in `size` middleware runs before `flip` and reports `availableWidth` and `availableHeight` for the pre-flip placement. After the boundary-aware flip changes placement, the available dimensions need recalculating.

### Suggested fix

In the overlay middleware pipeline, pass the reference element’s overflow ancestors as the boundary to `flip()`:

```tsx
if (this.config.flip !== false) {
	const boundary = getOverflowAncestors(this.trigger)
		.filter(a => a instanceof Element);

	middleware.push(flip(boundary.length ? { boundary } : undefined));
}
```
**No impact on the common case**

- If there is no scroll container, `getOverflowAncestors()` returns an empty array (or only Window or Document, which are filtered out).
- `flip()` is called with `undefined`, so behavior is identical to today.

Improves behavior inside scroll containers
- When the trigger is inside a scrollable ancestor, `flip()` receives those ancestors as `boundary`.
- It can then flip based on the *actual* available space, which matches user expectations.

### Steps to Reproduce

**Steps to reproduce**

1. Place a popover trigger inside a scrollable container, including one taking 100% of the viewport.
2. Scroll so the trigger is near the bottom of the container.
3. Open the popover with `placement: 'bottom'`.

**Expected**

- The popover flips to the top because there is no room below within the scroll container.

**Actual**

- The popover stays below the trigger and overflows the scroll container.
- Because the popover is portalled to `<body>`, Floating UI resolves clipping ancestors from the popover’s DOM position, not the trigger’s.
- It never discovers the trigger’s scroll container as a clipping boundary, regardless of the container’s size.

Image

### Reproduction

_No response_

gourlaouenp:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Documentation changes

## Issue

Closes #455
Closes #689

### Scroll blocking on ancestor containers (#455)

Rewrites `BlockScrollStrategy` to also block scrolling on ancestor scrollable containers, not just the document. Uses `overflow: hidden` + `scrollbarGutter: stable` instead of the CDK's `position: fixed` approach, which avoids the need to save/restore scroll position and prevents layout shift when the scrollbar disappears. Original inline styles are restored on disable.

### Flip and shift boundary options (#689)

Adds `boundary`, `rootBoundary`, and `crossAxis` options to `NgpFlipOptions` and `NgpShiftOptions`. The `flip` input type changes from `boolean` to `boolean | NgpFlipOptions` on primitives that previously only accepted a boolean (combobox, navigation-menu, popover, select, tooltip). This allows consumers to configure how Floating UI's flip and shift middleware detect overflow boundaries.

Updated documentation for combobox, menu, navigation-menu, popover, select, and tooltip.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/ng-primitives/portal/src/overlay.ts
+++ b/packages/ng-primitives/portal/src/overlay.ts
@@ -1,5 +1,4 @@
 import { FocusMonitor, FocusOrigin } from '@angular/cdk/a11y';
-import { ViewportRuler } from '@angular/cdk/overlay';
 import { DOCUMENT } from '@angular/common';
 import {
   DestroyRef,
@@ -228,7 +227,6 @@ export class NgpOverlay<T = unknown> implements CooldownOverlay {
   private readonly document = inject(DOCUMENT);
   private readonly destroyRef = inject(DestroyRef);
   private readonly viewContainerRef: ViewContainerRef;
-  private readonly viewportRuler = inject(ViewportRuler);
   private readonly focusMonitor = inject(FocusMonitor);
   private readonly cooldownManager = inject(NgpOverlayCooldownManager);
   /** Access any parent overlays */
@@ -715,7 +713,7 @@ export class NgpOverlay<T = unknown> implements CooldownOverlay {
   private createScrollStrategy(): ScrollStrategy {
     switch (this.config.scrollBehaviour) {
       case 'block':
-        return new BlockScrollStrategy(this.viewportRuler, this.document);
+        return new BlockScrollStrategy(this.document, this.config.triggerElement);
       case 'close':
         return new CloseScrollStrategy(
           this.config.anchorElement || this.config.triggerElement,
--- a/packages/ng-primitives/portal/src/scroll-strategy.ts
+++ b/packages/ng-primitives/portal/src/scroll-strategy.ts
@@ -1,105 +1,68 @@
 /**
- * This code is largely based on the CDK Overlay's scroll strategy implementation, however it
- * has been modified so that it does not rely on the CDK's global overlay styles.
+ * Scroll blocking inspired by react-aria's usePreventScroll.
+ * Uses `overflow: hidden` + `scrollbar-gutter: stable` instead of the CDK's
+ * `position: fixed` approach, avoiding the need to save/restore scroll position.
+ *
+ * @see https://github.com/adobe/react-spectrum/blob/main/packages/@react-aria/overlays/src/usePreventScroll.ts
  */
-import { coerceCssPixelValue } from '@angular/cdk/coercion';
-import { ViewportRuler } from '@angular/cdk/overlay';
 import { getOverflowAncestors } from '@floating-ui/dom';
-import { isFunction, isObject } from 'ng-primitives/utils';
 
 export interface ScrollStrategy {
   enable(): void;
   disable(): void;
 }
 
-/** Cached result of the check that indicates whether the browser supports scroll behaviors. */
-let scrollBehaviorSupported: boolean | undefined;
-
-export function supportsScrollBehavior(): boolean {
-  if (scrollBehaviorSupported != null) {
-    return scrollBehaviorSupported;
-  }
-
-  // If we're not in the browser, it can't be supported. Also check for `Element`, because
-  // some projects stub out the global `document` during SSR which can throw us off.
-  if (!isObject(document) || !document || !isFunction(Element) || !Element) {
-    return (scrollBehaviorSupported = false);
-  }
-
-  // If the element can have a `scrollBehavior` style, we can be sure that it's supported.
-  if ('scrollBehavior' in document.documentElement!.style) {
-    return (scrollBehaviorSupported = true);
-  }
-
-  // Check if scrollTo is supported and if it's been polyfilled
-  const scrollToFunction = Element.prototype.scrollTo;
-  if (!scrollToFunction) {
-    return (scrollBehaviorSupported = false);
-  }
-
-  // We can detect if the function has been polyfilled by calling `toString` on it. Native
-  // functions are obfuscated using `[native code]`, whereas if it was overwritten we'd get
-  // the actual function source. Via https://davidwalsh.name/detect-native-function. Consider
-  // polyfilled functions as supporting scroll behavior.
-  return (scrollBehaviorSupported = !/\{\s*\[native code\]\s*\}/.test(scrollToFunction.toString()));
-}
-
-interface HTMLStyles {
-  top: string;
-  left: string;
-  position: string;
+interface SavedOverflowState {
+  element: HTMLElement;
+  overflow: string;
+  overflowX: string;
   overflowY: string;
-  width: string;
+  scrollbarGutter: string;
 }
 
 export class BlockScrollStrategy implements ScrollStrategy {
-  private readonly previousHTMLStyles: HTMLStyles = {
-    top: '',
-    left: '',
-    position: '',
-    overflowY: '',
-    width: '',
-  };
-  private previousScrollPosition = { top: 0, left: 0 };
   private isEnabled = false;
+  private didBlockDocument = false;
+  private previousOverflow = '';
+  private previousScrollbarGutter = '';
+  private readonly savedAncestorStates: SavedOverflowState[] = [];
 
   constructor(
-    private readonly viewportRuler: ViewportRuler,
     private readonly document: Document,
+    private readonly triggerElement?: HTMLElement,
   ) {}
 
   /** Blocks page-level scroll while the attached overlay is open. */
   enable() {
-    if (this.canBeEnabled()) {
-      const root = this.document.documentElement!;
+    const root = this.document.documentElement!;
 
-      this.previousScrollPosition = this.viewportRuler.getViewportScrollPosition();
+    // Don't enable if already enabled or if another strategy (ours or CDK's) has blocked scrolling.
+    if (
+      this.isEnabled ||
+      root.hasAttribute('data-scrollblock') ||
+      root.classList.contains('cdk-global-scrollblock')
+    ) {
+      return;
+    }
 
-      // Cache the previous inline styles in case the user had set them.
-      this.previousHTMLStyles.left = root.style.left || '';
-      this.previousHTMLStyles.top = root.style.top || '';
-      this.previousHTMLStyles.position = root.style.position || '';
-      this.previousHTMLStyles.overflowY = root.style.overflowY || '';
-      this.previousHTMLStyles.width = root.style.width || '';
+    // Block scrollable ancestors of the trigger element regardless of document scrollability.
+    if (this.triggerElement) {
+      this.blockScrollableAncestors(this.triggerElement);
+    }
 
-      // Set the styles to block scrolling.
-      root.style.position = 'fixed';
+    // Block document scroll if the document itself is scrollable.
+    if (root.scrollHeight > root.clientHeight || root.scrollWidth > root.clientWidth) {
+      this.previousOverflow = root.style.overflow;
+      this.previousScrollbarGutter = root.style.scrollbarGutter;
 
-      // Necessary for the content not to lose its width. Note that we're using 100%, instead of
-      // 100vw, because 100vw includes the width plus the scrollbar, whereas 100% is the width
-      // that the element had before we made it `fixed`.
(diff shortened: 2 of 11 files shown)


packages/ng-primitives/portal/src/scroll-strategy.ts
```diff
@@ -107,51 +135,85 @@ export class BlockScrollStrategy implements ScrollStrategy {
...
-    // Since the scroll strategies can't be singletons, we have to use a global CSS class
-    // (`cdk-global-scrollblock`) to make sure that we don't try to disable global
-    // scrolling multiple times.
-    const html = this.document.documentElement!;
-
-    if (html.classList.contains('cdk-global-scrollblock') || this.isEnabled) {
-      return false;
+  private blockScrollableAncestors(element: HTMLElement): void {
+    let parent = element.parentElement;
+    while (parent && parent !== this.document.documentElement && parent !== this.document.body) {
```
ashley-hunter: Could this use the `getOverflowAncestors` here too instead of writing our own?

packages/ng-primitives/portal/src/overlay.ts
```diff
@@ -722,6 +726,17 @@ export class NgpOverlay<T = unknown> implements CooldownOverlay {
     // even when using programmatic position (the virtual element is created dynamically in computePosition)
     const referenceElement = this.config.anchorElement || this.config.triggerElement;
 
+    // Cache overflow ancestors once for boundary calculations (used by flip/shift/size).
+    // Only include ancestors that actually have scrollable content, not just those with
+    // overflow: auto/scroll CSS — otherwise non-scrollable containers over-constrain
+    // the flip boundary.
+    this.cachedOverflowAncestors = getOverflowAncestors(referenceElement).filter(
```
ashley-hunter: Do you foresee there being any issues with this change? I guess this is a change in behaviour? Do we need to make this opt-in? Or do you think this is a safe change to make?

packages/ng-primitives/portal/src/scroll-strategy.ts
```diff
@@ -107,51 +135,85 @@ export class BlockScrollStrategy implements ScrollStrategy {
...
+        // Disable user-defined smooth scrolling temporarily while we restore the scroll position.
+        // See https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior
+        // Note that we don't mutate the property if the browser doesn't support `scroll-behavior`,
+        // because it can throw off feature detections in `supportsScrollBehavior` which
+        // checks for `'scrollBehavior' in documentElement.style`.
+        if (scrollBehaviorSupported) {
+          htmlStyle.scrollBehavior = bodyStyle.scrollBehavior = 'auto';
+        }
+
+        window.scroll(this.previousScrollPosition.left, this.previousScrollPosition.top);
```
ashley-hunter: Do we need to store and restore the current scroll position of any other elements we have blocked?

ashley-hunter (commented):
Thanks for the fix! Just had a few questions.

packages/ng-primitives/portal/src/scroll-strategy.ts
gourlaouenp: Done

packages/ng-primitives/portal/src/overlay.ts
gourlaouenp: I killed it in favor of the DS composable by just giving access to flip options

packages/ng-primitives/portal/src/scroll-strategy.ts
gourlaouenp: Done

ashley-hunter (changes requested):
Thanks very much, I think the change to make it configurable is a good idea.

I did notice scroll blocking doesn't seem to work reliably on mobile. I haven't yet checked desktop, but if I scroll and scroll are are visible and open the menu sometimes I can still scroll the page which seems to be a regression.

https://github.com/user-attachments/assets/016bbc1f-f015-4b3d-9413-09e6454fd7b6

gourlaouenp:
> Thanks very much, I think the change to make it configurable is a good idea.
> 
> I did notice scroll blocking doesn't seem to work reliably on mobile. I haven't yet checked desktop, but if I scroll and scroll are are visible and open the menu sometimes I can still scroll the page which seems to be a regression.
> 
>  ScreenRecording_03-05-2026.22-20-07_1.mov

I don't have the issue on my phone, did you use this link? https://deploy-preview-690--ng-primitives.netlify.app/primitives/menu#menu

clementloridan (changes requested):
Looked at how [floating-ui `FloatingOverlay` lockScroll](https://floating-ui.com/docs/FloatingOverlay#lockscroll), [react-aria `usePreventScroll`](https://github.com/adobe/react-spectrum/blob/main/packages/@react-aria/overlays/src/usePreventScroll.ts) and [base-ui `useScrollLock`](https://github.com/mui/base-ui/blob/master/packages/react/src/utils/useScrollLock.ts) handle scroll locking, we're missing ref-counting (breaks when multiple components lock at once), and our scrollbar compensation needs work. iOS is its own mess.

packages/ng-primitives/portal/src/scroll-strategy.ts
```diff
@@ -1,158 +1,133 @@
...
  */
-import { coerceCssPixelValue } from '@angular/cdk/coercion';
-import { ViewportRuler } from '@angular/cdk/overlay';
 import { getOverflowAncestors } from '@floating-ui/dom';
-import { isFunction, isObject } from 'ng-primitives/utils';
 
 export interface ScrollStrategy {
   enable(): void;
   disable(): void;
 }
```
clementloridan, line 13: ### Missing ref counting proposal:

If a dialog opens a popover and one closes, it'll restore styles and unlock scroll while the other is still open. floating-ui, react-aria and base-ui all use a global lock count to handle this.
I'd extract a ScrollLocker singleton (like base-ui) with acquire() / release() instead of touching styles directly from BlockScrollStrategy:


```ts
class ScrollLocker {
  private static instance: ScrollLocker;
  private lockCount = 0;
  private previousOverflow = '';
  private previousScrollbarGutter = '';

  static getInstance() { ... }

  acquire(document: Document): void {
    this.lockCount++;
    if (this.lockCount === 1) { this.lock(document); }
  }

  release(document: Document): void {
    this.lockCount--;
    if (this.lockCount === 0) { this.unlock(document); }
  }
}
```

> base-ui also wraps acquire / release in setTimeout(0) to avoid flicker on simultaneous transitions, probably worth it.

packages/ng-primitives/portal/src/scroll-strategy.ts
```diff
@@ -1,158 +1,133 @@
...
-      this.previousHTMLStyles.position = root.style.position || '';
-      this.previousHTMLStyles.overflowY = root.style.overflowY || '';
-      this.previousHTMLStyles.width = root.style.width || '';
-
-      // Set the styles to block scrolling.
-      root.style.position = 'fixed';
+    const root = this.document.documentElement!;
+
+    // Don't enable if already enabled or if another strategy (ours or CDK's) has blocked scrolling.
+    if (
```
clementloridan, line 40: ### No iOS Safari handling proposal

iOS Safari ignores overflow: hidden on the document in several scenarios (collapsed address bar, keyboard visible, etc.), this is what @ashley-hunter  flagged in the earlier review.

Simplest fix is the `floating-ui` approach: detect iOS and switch to `position: fixed` on the body with scroll position offset:

```ts
const isIOS = /iP(hone|ad|od)|iOS/.test(navigator.platform)
  || (navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1);

if (isIOS) {
  const scrollX = window.scrollX;
  const scrollY = window.scrollY;
  Object.assign(root.style, {
    position: 'fixed',
    top: `${-scrollY}px`,
    left: `${-scrollX}px`,
    right: '0',
    overflow: 'hidden',
  });
  // On unlock: restore styles + window.scrollTo(scrollX, scrollY)
}
```

packages/ng-primitives/portal/src/scroll-strategy.ts
```diff
@@ -1,158 +1,133 @@
...
+      this.isEnabled ||
+      root.hasAttribute('data-scrollblock') ||
```
clementloridan, line 42: **`scrollbar-gutter: stable` has no fallback — causes visible layout shift on unsupported browsers.**

Ashley noticed this: "there just seems to be a white gap where the scroll bar was previously." This is the exact symptom of `scrollbar-gutter: stable` without a visible scrollbar to fill the gutter.

react-aria handles this with a runtime check + fallback:

```ts
const scrollbarWidth = window.innerWidth - document.documentElement.clientWidth;
if (scrollbarWidth > 0) {
  if ('scrollbarGutter' in document.documentElement.style) {
    root.style.scrollbarGutter = 'stable';
  } else {
    root.style.paddingRight = `${scrollbarWidth}px`;
  }
}
```

floating-ui goes further and exposes a CSS custom property so consumers can compensate their own fixed-position elements (e.g. sticky headers):

```ts
root.style.setProperty('--scrollbar-width', `${scrollbarWidth}px`);
```

**Suggestion:** Use the react-aria approach: detect `scrollbar-gutter` support at runtime and fall back to `padding-right`. Also consider exposing a CSS variable for consumer compensation.

ashley-hunter:
Updated branch and covered in #866

PR #574, opened 2025-12-21
refactor(select): state pattern migration 

zecka:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] The commit message follows our guidelines: https://github.com/ng-primitives/ng-primitives/blob/main/CONTRIBUTING.md#-commit-message-guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes
- [ ] Other... Please describe:

## Issue

Closes #

## Does this PR introduce a breaking change?

- [ ] Yes
- [ ] No

PR #692, opened 2026-03-04
refactor(select): state pattern migration

### LINKED ISSUE #559: Select State Pattern Migration
Migrate select component to new state pattern

zecka:
For context, I previously opened a PR (#574), but since many changes happened in the meantime, I decided to start over from scratch to make sure nothing was missed.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

Closes #559 

Implement state pattern on select component

- [ ] Yes
- [x] No

## Summary by CodeRabbit

* **New Features**
  * Controlled-signal utility plus reactive binding helpers for attributes, styles and data-* attributes.
  * New primitive-backed select pieces (native select, dropdown, option, portal) exposing stable props and lifecycle methods.

* **Refactor**
  * Consolidated select internals into a centralized, primitive-driven state for simpler wiring and more consistent behaviour while preserving public inputs/outputs.

* **Bug Fixes / Tests**
  * Improved disabled-attribute handling and test setup to clear scrollIntoView mocks.


zecka:
@ashley-hunter
I saw that you added `isomorphicEffect` in #578, but I still need to force `effect` instead of `afterRenderEffect`, for the same reason discussed in the previous PR (#574).

So I added three functions (`dataBindingEffect`, `styleBindingEffect`, and `attrBindingEffect`). I am not entirely sure this is the right approach, as this is only meant to fix test issues.

For now, I still have failing tests related to Virtual Scrolling. I will investigate this over the next few days (it also seems to be test-related)

packages/ng-primitives/select/src/select/select.spec.ts
```diff
@@ -1205,10 +1208,8 @@ describe('NgpSelect', () => {
         const { fixture } = await render(VirtualScrollingTestComponent);
         const component = fixture.componentInstance;
 
-        // Spy on scrollIntoView to ensure it's not called
-        const scrollIntoViewSpy = jest
-          .spyOn(Element.prototype, 'scrollIntoView')
-          .mockImplementation();
+        const scrollIntoViewMock = Element.prototype.scrollIntoView as jest.Mock;
+        scrollIntoViewMock.mockClear();
```
zecka: @ashley-hunter  I have to clear mock here, because test fail.
Because scollIntoView was trigger in previous test

Maybe we should add this into `packages/ng-primitives/src/test-setup.ts` ?

```ts
beforeEach(() => {
  (Element.prototype.scrollIntoView as jest.Mock).mockClear();
});

```

packages/ng-primitives/state/src/index.ts
```diff
@@ -368,6 +368,19 @@ export function createPrimitive<TFactory extends (...args: any[]) => unknown>(
...
+ * This utility serves two purposes:
+ * 1. **Synchronization**: Keeps internal state synced when external signal changes
+ * 2. **API Safety**: When returned with `.asReadonly()`, guarantees immutability in public API
+ *
+ * Use `controlled()` for all props to ensure consistent readonly API surface,
+ * even if the prop isn't modified internally.
+ *
+ * @param value - The external signal to synchronize with
+ * @returns A writable signal linked to the external signal
+ */
```
zecka, line 386: I had trouble understanding the purpose of this function, so I added my understanding in the JSDoc. Please confirm that this description is correct.

packages/ng-primitives/select/src/select/select.spec.ts
ashley-hunter: Yes, might be a good idea!

ashley-hunter:
Thank you, I will look into the effect issue you were seeing see if I can come up with anything. 
I'll also test it out in the design system I work on, as select is a very widely used component want to make sure we don't have any subtle breakages 🙂 appreciate the great work!

packages/ng-primitives/select/src/select/select.ts
```diff
@@ -131,479 +97,19 @@ export class NgpSelect {
...
-  }
-
-  /**
-   * Open the dropdown.
-   * @internal
-   */
-  async openDropdown(): Promise<void> {
-    if (this.state.disabled() || this.open()) {
-      return;
-    }
```
zecka, line 237: @ashley-hunter  I was thinking about something.
Before the state-pattern migration, there were some internal methods inside the directive that allowed opening the dropdown programmatically (even if it was tagged `@internal`, people maybe use it).

Something like:

```ts
@Component({
  template: `
    <div ngpSelect>
      <div *ngpSelectPortal ngpSelectDropdown>
          ....
      </div>
    </div>
`
})
export class AppFoo {
   select = viewChild(NgpSelect);
   
   openSelectDropdown(){
     this.select().openDropdown()
   }
}

```
So now, is there no way anymore to open the select dropdown programmatically?
Should I expose some kind of bridge method inside the directive that forwards to the state method? 

Something like:
```ts
openDropdown(){
  return this.state.openDropdown()
}
```

ashley-hunter: Yes absolutely, this is what we have done with all the other migrations. The public API of the directives technically shouldn't change at all, the method just wouldn't actually perform the action. It would delegate it to the state as you have suggested.

zecka: Okay so even if the method is tagged `@internal` we should keep it on the directive too ?

zecka: Maybe this is something i miss on previous migration
For example:
https://github.com/ng-primitives/ng-primitives/pull/570/changes#diff-409bb9208a8906089645a49e95a41adbf775a069e23b2b0808ee0d7e1f3f5f8aL159-L169

I can make a fix for this one, just told me 😊

zecka: I have restored the public APIs.

packages/ng-primitives/state/src/index.ts
```diff
@@ -457,6 +507,30 @@ export function styleBinding(
   });
 }
 
+export function dataBindingEffect(
```
ashley-hunter: Finally got some time to look into this - I think we can remove these functions and use the existing ones, for the tests we can just use the waitFor helper which will avoid the failures, in reality this seems to work fine in the browser using the existing functions

zecka: Okay make sense, I will update the PR

packages/ng-primitives/select/src/select-option/select-option-state.ts
```diff
@@ -0,0 +1,221 @@
...
+      if (disabled()) {
+        return;
+      }
+
+      onActivated?.();
+      selectState().toggleOption(id());
+    }
+
+    function scrollIntoView(): void {
+      elementRef.nativeElement.scrollIntoView({ block: 'nearest' });
```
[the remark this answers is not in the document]
ashley-hunter: I think this issue is legitimate, I'm seeing cases where the page scrolls to the bottom when opening select

ashley-hunter (changes requested):
Thanks, looks great! I notice an issue where the page is scrolling to the bottom on select open, and also am noticing a bit of flickering when opening the select.

zecka:
> Thanks, looks great! I notice an issue where the page is scrolling to the bottom on select open, and also am noticing a bit of flickering when opening the select.

Good catch @ashley-hunter. I had actually noticed this issue earlier and it was previously fixed when `styleBindingEffect` was introduced. After reverting `styleBindingEffect`, I re-tested but didn’t run this particular smoke test again.

The page scroll happens because of a timing issue in `openDropdown` in `select-state`. The function calls `await portal()?.show()`. At that moment the dropdown is inserted into the DOM, but `styleBinding` hasn’t finished running yet since it relies on `afterRenderEffect` rather than `effect`. As a result, the dropdown is still visually positioned at the bottom of the page, so `scrollTo` scrolls there.

To fix this I added the following after opening the dropdown:

```ts
await new Promise<void>(resolve => {
  runInInjectionContext(injector, () => {
    afterRenderEffect(() => resolve());
  });
});
```
I’m wondering whether this wait should live inside `select-portal-state.show`, or even `NgpOverlay.show`, so callers don't need to handle this timing.

Regarding the flickering, I’m not sure. I do notice a small flash when the component initially renders, but that already seems to happen before this PR. Is that what you’re referring to?

zecka:
I add a new commit to make a small fix on resize utilities. The goal of is to emit resize dimensions immediately on subscription, instead of waiting for the first ResizeObserver callback
.
After that change, I noticed a failing popover test caused by an extra close event from the initial resize emission, so I also updated `overlay.ts` to run resize-driven hide logic only when the overlay is actually open. 

Does this look OK to you?

packages/ng-primitives/portal/src/overlay.ts
```diff
@@ -590,6 +590,11 @@ export class NgpOverlay<T = unknown> implements CooldownOverlay {
       this.closeTimeout = undefined;
```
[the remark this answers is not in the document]
zecka: Hmm... fixing it would require tracking Promise resolvers. Maybe it adding complexity for an unlikely scenario that doesn't have a lot of probability to manifest in normal usage.
If I understand well the scenario can appear if the trigger element gets hidden (resized to 0x0) while show() is still pending during showDelay.

ashley-hunter:
@cubic-dev-ai

ashley-hunter:
@cubic-dev-ai review

ashley-hunter (approved):
Thank you for all your hard work and patience - apologies for the delay - looks great and works well.

PR #704, opened 2026-03-14
feat(dialog)!: replace CDK overlay with standalone portal

clementloridan:
## Summary

- Re-implement dialog overlay using `ng-primitives/portal` instead of `@angular/cdk/overlay`
- Introduce `NgpOverlayRegistry` for centralized escape-key routing and parent-child overlay tracking
- Update dialog documentation with full `NgpDialogManager` and `NgpDialogRef` API reference
- Clean up stale CDK references in comments and fix minor typo

## Test plan

- [x] `pnpm build` passes
- [x] `nx test ng-primitives` — all 1120 tests pass
- [ ] Visual check: `pnpm start` → navigate to dialog docs page, verify rendering
- [ ] Verify dialog open/close, escape key, overlay click, and navigation close behaviors

---
## Summary by cubic
Replaced the dialog overlay with a standalone portal from `ng-primitives/portal`, removing `@angular/cdk/overlay`. Added a centralized overlay registry, expanded dialog APIs, and CDK-compatible outside click/Escape behavior.

- **New Features**
  - Central `NgpOverlayRegistry` for Escape routing, topmost-only dismiss, and parent–child cascade close; auto-detects parent overlays; exports `NgpOverlayRef` and registry types.
  - `NgpDialogRef` now implements `NgpOverlayRef`; adds `afterClosed`, `keydownEvents`, and `outsidePointerEvents` (via registry), plus a deprecated `overlayRef` shim. `NgpDialogManager` adds `getDialogById`, `openDialogs`, `afterOpened`, `afterAllClosed`.
  - Standalone `BlockScrollStrategy` using `@angular/cdk/scrolling`’s `ViewportRuler`, honoring both `data-scrollblock` and `cdk-global-scrollblock`; hides non-dialog content from AT; dialogs close on browser and programmatic navigation.

- **Bug Fixes**
  - Reset scroll blocking after the last dialog closes so each open uses the correct strategy.
  - Cascade teardown ignores `disableClose`, preventing orphaned child dialogs when a parent closes.
  - `outsidePointerEvents` now comes from the registry using CDK’s pointerdown+click pattern with stacking and Shadow DOM support; examples set `z-index: 1000` to avoid overlay ordering issues.

<sup>Written for commit 82fcc5f80535b0669fed0efee6cadbe20d7e7fcd. Summary will update on new commits.</sup>

### DIFF
--- a/apps/documentation/src/app/examples/dialog/dialog-drawer.example.ts
+++ b/apps/documentation/src/app/examples/dialog/dialog-drawer.example.ts
@@ -67,6 +67,7 @@ import {
       backdrop-filter: blur(4px);
       position: fixed;
       inset: 0;
+      z-index: 1000;
       display: flex;
       justify-content: flex-end;
       align-items: stretch;
--- a/apps/documentation/src/app/pages/(documentation)/primitives/dialog.md
+++ b/apps/documentation/src/app/pages/(documentation)/primitives/dialog.md
@@ -111,7 +111,7 @@ The following directives are available to import from the `ng-primitives/dialog`
 
 The `NgpDialogManager` can be used as an alternative to the `NgpDialogTrigger` directive for programmatically opening dialogs.
 
-The manager provides a method to open or close all dialogs and accepts a component or template reference to display.
+The manager provides methods to open, close, and query dialogs, and accepts a component or template reference to display.
 
 <prop-details name="open" type="(component: Type | TemplateRef, options?: NgpDialogConfig) => NgpDialogRef">
   Opens a dialog with the specified component or template reference.
@@ -119,6 +119,58 @@ The manager provides a method to open or close all dialogs and accepts a compone
 
 <prop-details name="closeAll" type="() => void">
   Closes all open dialogs.
+</prop-details>
+
+<prop-details name="getDialogById" type="(id: string) => NgpDialogRef | undefined">
+  Finds an open dialog by its ID.
+</prop-details>
+
+<prop-details name="openDialogs" type="readonly NgpDialogRef[]">
+  The list of currently open dialogs.
+</prop-details>
+
+<prop-details name="afterOpened" type="Subject<NgpDialogRef>">
+  Stream that emits when a dialog has been opened.
+</prop-details>
+
+<prop-details name="afterAllClosed" type="Observable<void>">
+  Stream that emits when all open dialogs have finished closing. Emits on subscribe if there are no open dialogs.
+</prop-details>
+
+### NgpDialogRef
+
+Reference returned by `NgpDialogManager.open()`. Provides methods to interact with the opened dialog.
+
+<prop-details name="close" type="(result?: R, focusOrigin?: FocusOrigin) => Promise<void>">
+  Closes the dialog, optionally returning a result value.
+</prop-details>
+
+<prop-details name="afterClosed" type="Observable<R | undefined>">
+  Observable that emits the dialog result when the dialog is closed.
+</prop-details>
+
+<prop-details name="data" type="T">
+  The data passed to the dialog via `NgpDialogConfig.data`.
+</prop-details>
+
+<prop-details name="id" type="string">
+  The unique ID for the dialog.
+</prop-details>
+
+<prop-details name="disableClose" type="boolean | undefined">
+  Whether the user is allowed to close the dialog.
+</prop-details>
+
+<prop-details name="keydownEvents" type="Observable<KeyboardEvent>">
+  Observable that emits keyboard events dispatched within the dialog.
+</prop-details>
+
+<prop-details name="outsidePointerEvents" type="Observable<MouseEvent>">
+  Observable that emits pointer events dispatched outside of the dialog.
+</prop-details>
+
+<prop-details name="updatePosition" type="() => NgpDialogRef">
+  Updates the position of the dialog. Currently a no-op as dialogs are CSS-centered.
 </prop-details>
 
 ## Examples
--- a/packages/ng-primitives/dialog/src/dialog/dialog-ref.ts
+++ b/packages/ng-primitives/dialog/src/dialog/dialog-ref.ts
@@ -48,8 +48,15 @@ export class NgpDialogRef<T = unknown, R = unknown> implements NgpOverlayRef {
   /** Emits on keyboard events within the dialog. */
   readonly keydownEvents: Observable<KeyboardEvent>;
 
+  /**
+   * Emits pointer events (click, auxclick, contextmenu) that happen outside of the dialog.
+   * Fed by the NgpOverlayRegistry with CDK-compatible stacking awareness.
+   * @internal
+   */
+  readonly outsidePointerEvents$ = new Subject<MouseEvent>();
+
   /** Emits on pointer events that happen outside of the dialog. */
-  readonly outsidePointerEvents: Observable<MouseEvent>;
+  readonly outsidePointerEvents: Observable<MouseEvent> = this.outsidePointerEvents$.asObservable();
 
   constructor(
     readonly config: NgpDialogConfig<T>,
@@ -69,14 +76,6 @@ export class NgpDialogRef<T = unknown, R = unknown> implements NgpOverlayRef {
       );
     });
 
-    this.outsidePointerEvents = defer(() => {
-      const elements = this.getElements();
-      if (!elements.length) return EMPTY;
-      return fromEvent<MouseEvent>(this.document, 'mouseup').pipe(
-        filter(event => !elements.some(el => el.contains(event.target as Node))),
-        takeUntil(this.closed),
-      );
-    });
   }
 
   /**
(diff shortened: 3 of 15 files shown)


packages/ng-primitives/portal/src/overlay-registry.ts
```diff
@@ -0,0 +1,332 @@
...
+    if (event.key !== 'Escape' || event.isComposing) {
+      return;
+    }
+
+    const topmost = this.getTopmost();
+    if (!topmost || this.pendingGuardIds.has(topmost.id)) {
+      return;
+    }
+
+    this.evaluateGuardAndDismiss(topmost.id, topmost.dismissPolicy.escapeKey, event, () =>
```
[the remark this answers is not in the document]
clementloridan: `stopPropagation` on Escape key in the overlay registry was investigated but not applied. Several components (e.g. navigation-menu) register `closeOnEscape` with the registry but also have their own Escape handlers in bubble phase for focus restoration. Since the registry listens in capture phase, `stopPropagation` would prevent those handlers from running, breaking focus restore. A proper fix requires unifying Escape handling so components rely solely on the registry callback rather than redundant keydown listeners.

packages/ng-primitives/portal/src/scroll-strategy.ts
```diff
@@ -128,31 +124,27 @@ export class BlockScrollStrategy implements ScrollStrategy {
...
   }
 
   private canBeEnabled(): boolean {
-    // Since the scroll strategies can't be singletons, we have to use a global CSS class
-    // (`cdk-global-scrollblock`) to make sure that we don't try to disable global
-    // scrolling multiple times.
     const html = this.document.documentElement!;
 
-    if (html.classList.contains('cdk-global-scrollblock') || this.isEnabled) {
+    if (html.hasAttribute('data-scrollblock') || this.isEnabled) {
```
ashley-hunter: We should keep the CDK selector here too, it was added to help integrate well with existing apps mixing CDK and ng-primitives.

packages/ng-primitives/portal/src/scroll-strategy.ts
```diff
@@ -63,17 +62,14 @@ export class BlockScrollStrategy implements ScrollStrategy {
...
-  ) {}
+  constructor(private readonly document: Document) {}
 
   /** Blocks page-level scroll while the attached overlay is open. */
   enable() {
     if (this.canBeEnabled()) {
       const root = this.document.documentElement!;
 
-      this.previousScrollPosition = this.viewportRuler.getViewportScrollPosition();
+      this.previousScrollPosition = { top: window.scrollY ?? 0, left: window.scrollX ?? 0 };
```
ashley-hunter: Can we continue to use the viewport ruler, we are keeping the CDK dependency, just moving away from their dialog, so this would be one less place to risk introducing regressions.

packages/ng-primitives/dialog/src/dialog/dialog-ref.ts
```diff
@@ -19,56 +25,102 @@ export class NgpDialogRef<T = unknown, R = unknown> {
...
+        filter(event => elements.some(el => el.contains(event.target as Node))),
+        takeUntil(this.closed),
+      );
     });
 
-    this.detachSubscription = overlayRef.detachments().subscribe(() => this.close());
+    this.outsidePointerEvents = defer(() => {
+      const elements = this.getElements();
+      if (!elements.length) return EMPTY;
+      return fromEvent<MouseEvent>(this.document, 'mouseup').pipe(
```
ashley-hunter: I think this is slightly different than the CDK behavior which is an observable from OverlayRef that emits any pointer interaction (mouse/touch/pointer) that occurs outside the overlay pane, with built-in overlay stacking awareness.

apps/documentation/src/app/pages/(documentation)/primitives/dialog.md
```diff
@@ -111,7 +111,7 @@ The following directives are available to import from the `ng-primitives/dialog`
 
 The `NgpDialogManager` can be used as an alternative to the `NgpDialogTrigger` directive for programmatically opening dialogs.
```
ashley-hunter, line 112: image

the drawer example in the docs seems broken. Ideally we should try and fix this in the implementation rather than changing the example to avoid any breaking changes for consumers

ashley-hunter (changes requested):
Thank you! Had a quick review of the changes, and had a few small suggestions.

packages/ng-primitives/dialog/src/dialog/dialog-ref.ts
clementloridan: Good point, our `outsidePointerEvents` on `NgpDialogRef` currently only listens to `mouseup`, whereas the CDK covers `pointerdown`, `click`, `auxclick`, and `contextmenu`.

However, the actual dismiss logic is now handled by `NgpOverlayRegistry.handleOutsideClick`, which already has stacking awareness, dismiss guards, scrollbar detection, and shadow DOM support.                                                                                                                                                                           

The `outsidePointerEvents` observable on `NgpDialogRef` is only exposed through the deprecated overlayRef shim. Should we:

1. Expand it to match CDK event types (mouse + touch + pointer) for better compat during the deprecation period?
2. Leave it as-is since the real dismiss path goes through the registry?

Happy to go either way.

apps/documentation/src/app/pages/(documentation)/primitives/dialog.md
clementloridan: Looking into this: the drawer example uses `position: fixed; inset: 0` on the overlay, which is correct, but without the CDK's `.cdk-overlay-container` (which provided position: fixed; z-index: 1000 globally), the portal-rendered elements can end up behind other stacking contexts on the page (navbar, sidebar, etc.).

I see two approaches to fix this in the implementation:

A) Shared overlay container: Create a global `ngp-overlay-container` element (`position: fixed; inset: 0; z-index: 1000; pointer-events: none`) and attach all portals inside it, similar to what the CDK did.

B) Inline z-index on portal elements — Have the portal set a high `z-index` directly on the root elements it projects into the body.

Option A feels cleaner and more predictable for stacking multiple overlays. Do you have a preference, or a different approach in mind?

packages/ng-primitives/dialog/src/dialog/dialog-ref.ts
ashley-hunter: I think the safest approach might be to replicate the existing functionality, just to avoid introducing any breaking changes in existing usages

apps/documentation/src/app/pages/(documentation)/primitives/dialog.md
ashley-hunter: Hmm, this is a tricky one.. introducing a container for all overlays could be a breaking change. Adding a container for just dialogs could affect how other overlays work inside a dialog 🤔 - setting a z-index on all overlays could also cause breaking changes too. Bit of a tricky position. I think in this case the least invasive fix would be to add z-index to the drawer example code. Any thoughts?

packages/ng-primitives/portal/src/scroll-strategy.ts
clementloridan: Done in fe077db3 — restored the `cdk-global-scrollblock` selector check alongside our `data-scrollblock` attribute.

packages/ng-primitives/portal/src/scroll-strategy.ts
clementloridan: Done in fe077db3 — restored ViewportRuler usage throughout the scroll strategy.

packages/ng-primitives/dialog/src/dialog/dialog-ref.ts
clementloridan: Fixed in 329ee8fd — `outsidePointerEvents` is now a Subject fed by the `NgpOverlayRegistry` using CDK's pattern: tracks `pointerdown` origin then listens to `click`, `auxclick`, and `contextmenu` in capture phase, with stacking awareness (iterates overlays topmost-to-oldest, stops when one contains the click) and shadow DOM support via `composedPath()`.

apps/documentation/src/app/pages/(documentation)/primitives/dialog.md
clementloridan: Agreed — added `z-index: 1000` to the drawer overlay in both CSS and Tailwind examples in 329ee8fd. Keeps it non-breaking for consumers.

apps/documentation/src/app/examples/dialog/dialog-drawer.example.ts
```diff
@@ -67,6 +67,7 @@ import {
       backdrop-filter: blur(4px);
       position: fixed;
       inset: 0;
+      z-index: 1000;
```
ashley-hunter, line 70: image

Looks like we also need to do this for the other dialog examples too.

ashley-hunter (changes requested):
Thanks, everything looks goes and seems to work well. I just noticed the other dialog examples in the docs need the z index change too, then I think this is good to merge 😊

apps/documentation/src/app/examples/dialog/dialog-drawer.example.ts
clementloridan: I fixed!

ashley-hunter (approved):
Great work, thank you!

ashley-hunter:
@allcontributors add @clementloridan for code

PR #719, opened 2026-03-30
feat(dialog,popover): dismiss guards for closeOnEscape and closeOnOutsideClick

clementloridan:
## Summary
- Add dismiss guard support (`closeOnEscape`, `closeOnOutsideClick`) to dialog and popover, allowing sync or async guard functions (including Promises) to conditionally prevent dismissal
- Add missing `closeOnOutsideClick` input to `NgpDialogTrigger`
- Split `closed` (emits immediately) and `afterClosed` (emits after exit animations) on `NgpDialogRef`, aligned to same `{ focusOrigin?, result? }` shape
- Popover dismiss guard examples now close in parallel with confirmation dialog instead of sequentially

## Test plan
- [ ] Dialog dismiss guard example: escape and outside click respect the guard
- [ ] Popover dismiss guard example: popover closes simultaneously with confirmation dialog
- [ ] Dialog examples without guards still work as before
- [ ] `afterClosed` emits after exit animations complete
- [ ] `closed` emits immediately when `close()` is called

---
## Summary by cubic
Adds dismiss guards to `ng-primitives/dialog` and `ng-primitives/popover` so `closeOnEscape` and `closeOnOutsideClick` accept booleans or guard functions (sync or async). Also adds `[ngpDialogTriggerCloseOnOutsideClick]`, splits `NgpDialogRef.closed` (immediate) from `afterClosed` (post-animation), updates docs/examples, and restores focus after exit.

- **New Features**
  - `closeOnEscape`/`closeOnOutsideClick` accept `NgpDismissGuard` (`boolean | (event) => boolean | Promise<boolean>`); string boolean attributes are coerced via `dismissGuardAttribute` from `ng-primitives/portal`.
  - `NgpDialogTrigger` supports `[ngpDialogTriggerCloseOnEscape]` and `[ngpDialogTriggerCloseOnOutsideClick]` with guard functions; overlay respects deprecated `closeOnClick` when no guard is set.
  - `NgpDialogRef` exposes `closed` (emits immediately) and keeps `afterClosed: Observable<R | undefined>` (emits after exit); focus restores after exit using `focusOrigin`.
  - Docs and examples updated for `ng-primitives/dialog` and `ng-primitives/popover`; popover guard closes in parallel with the confirm dialog and resets dirty state on discard; tests use `jest.fn()`, add `waitFor`, and use `dispatchOverlayClick`.

- **Migration**
  - Replace deprecated `closeOnClick` with `closeOnOutsideClick` in dialog config.
  - Use `closed` to react immediately; keep using `afterClosed` for results after animations.
  - For conditional closes, pass a function to `closeOnEscape`/`closeOnOutsideClick` that returns `boolean | Promise<boolean>`.

<sup>Written for commit ce04150ce5e7b65a70bd30128168397c3f975dd7. Summary will update on new commits.</sup>

### DIFF
--- a/packages/ng-primitives/dialog/src/dialog/dialog-ref.ts
+++ b/packages/ng-primitives/dialog/src/dialog/dialog-ref.ts
@@ -3,7 +3,7 @@ import { inject, Injector } from '@angular/core';
 import { NgpExitAnimationManager } from 'ng-primitives/internal';
 import { NgpDismissGuard, NgpOverlayRef } from 'ng-primitives/portal';
 import { defer, EMPTY, fromEvent, Observable, Subject } from 'rxjs';
-import { filter, takeUntil } from 'rxjs/operators';
+import { filter, map, takeUntil } from 'rxjs/operators';
 import { NgpDialogConfig } from '../config/dialog-config';
 
 /** Minimal portal interface needed by the dialog ref. */
@@ -29,13 +29,14 @@ export class NgpDialogRef<T = unknown, R = unknown> implements NgpOverlayRef {
   readonly closed = new Subject<{ focusOrigin?: FocusOrigin; result?: R }>();
 
   /** @internal */
-  private readonly afterClosed$ = new Subject<{ focusOrigin?: FocusOrigin; result?: R }>();
+  readonly afterClosed$ = new Subject<{ focusOrigin?: FocusOrigin; result?: R }>();
 
   /**
    * Observable that emits the dialog result after exit animations have completed.
    */
-  readonly afterClosed: Observable<{ focusOrigin?: FocusOrigin; result?: R }> =
-    this.afterClosed$.asObservable();
+  readonly afterClosed: Observable<R | undefined> = this.afterClosed$.pipe(
+    map(event => event.result),
+  );
 
   /** Data passed from the dialog opener. */
   readonly data: T;
--- a/packages/ng-primitives/dialog/src/dialog/dialog-ref.ts
+++ b/packages/ng-primitives/dialog/src/dialog/dialog-ref.ts
@@ -3,7 +3,7 @@ import { inject, Injector } from '@angular/core';
 import { NgpExitAnimationManager } from 'ng-primitives/internal';
 import { NgpDismissGuard, NgpOverlayRef } from 'ng-primitives/portal';
 import { defer, EMPTY, fromEvent, Observable, Subject } from 'rxjs';
-import { filter, map, takeUntil } from 'rxjs/operators';
+import { filter, takeUntil } from 'rxjs/operators';
 import { NgpDialogConfig } from '../config/dialog-config';
 
 /** Minimal portal interface needed by the dialog ref. */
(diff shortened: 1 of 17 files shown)


packages/ng-primitives/popover/src/popover-trigger/popover-trigger.ts
```diff
@@ -150,21 +151,19 @@ export class NgpPopoverTrigger<T = null> implements OnDestroy {
...
-    transform: booleanAttribute,
   });
 
   /**
-   * Define whether the popover should close when the escape key is pressed.
+   * Define whether the popover should close when the escape key is pressed, or a guard function.
    * @default true
    */
-  readonly closeOnEscape = input<boolean, BooleanInput>(this.config.closeOnEscape, {
+  readonly closeOnEscape = input<NgpDismissGuard<KeyboardEvent>>(this.config.closeOnEscape, {
```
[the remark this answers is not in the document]
clementloridan: This is by design, not a bug. The input type was intentionally changed from `boolean` to `NgpDismissGuard<KeyboardEvent> (boolean | ((event) => boolean | Promise<boolean>))` to support dismiss guard functions. Adding `booleanAttribute` coercion would prevent passing a guard function, which is the whole point of this feature. These inputs are always used with property binding `([closeOnEscape]="value")`, not as plain HTML attributes, so string coercion is not a concern.

packages/ng-primitives/popover/src/popover-trigger/popover-trigger.ts
```diff
@@ -150,21 +151,19 @@ export class NgpPopoverTrigger<T = null> implements OnDestroy {
...
     alias: 'ngpPopoverTriggerCloseOnEscape',
-    transform: booleanAttribute,
```
ashley-hunter, line 167: I guess this is technically a breaking change, as we previously allowed static string Boolean values. I think perhaps we should add in our own transform that handles promises and if not a promise fallback to booleanAttribute or something similar to avoid the breaking change.

packages/ng-primitives/dialog/src/dialog/dialog-ref.ts
```diff
@@ -19,16 +19,23 @@ export class NgpDialogRef<T = unknown, R = unknown> implements NgpOverlayRef {
...
+  closeOnEscape: NgpDismissGuard<KeyboardEvent> | undefined;
+
+  /** Whether clicking outside (on the overlay) is allowed to close the dialog, or a guard function. */
+  closeOnOutsideClick: NgpDismissGuard<Element> | undefined;
 
   /** Emits when the dialog has been closed. */
   readonly closed = new Subject<{ focusOrigin?: FocusOrigin; result?: R }>();
 
+  /** @internal */
+  private readonly afterClosed$ = new Subject<{ focusOrigin?: FocusOrigin; result?: R }>();
```
ashley-hunter: I guess this is a breaking change - was the change in type necessary?

ashley-hunter (changes requested):
Thank you, looks good, just 2 small comments/questions

ashley-hunter:
Apologies for the delay - changes look good, just a merge conflict then this should be good to go, thanks!

ashley-hunter (approved):

PR #771, opened 2026-06-03
fix(state): apply host bindings during change detection to avoid first-paint flash

ashley-hunter:
## PR Checklist

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added
- [ ] Docs have been added / updated — N/A (internal binding-timing fix; no public API or example surface)

## PR Type

- [x] Bugfix

Host bindings (`attrBinding`/`dataBinding`/`styleBinding`) were applied via `afterRenderEffect`, which defers the first application to the frame after the element is created. In a zoneless app, client-side navigation painted the element in its default state for one frame before the binding landed - e.g. an initially-disabled button flashing from its enabled colour into the disabled colour.

`bindingEffect` now also applies the initial value synchronously during change detection (a dependency-free `untracked` effect), so the first paint is already correct; `afterRenderEffect` still handles reactive updates. SSR is unaffected (`afterRenderEffect` is a no-op on the server, so the synchronous effect ships the value).

Because `tabindex` now lands during change detection, roving-focus items register synchronously so the active item is tabbable before a focus trap focuses it on open - the focus trap itself is unchanged. Toggle-group coerces a null value (forms reset) to an empty array so the synchronous read stays safe.

Tests added: synchronous disabled binding (button), roving-focus active tabindex during CD, and a focus-trap focusing a roving-focus group.

- [x] No

---
## Summary by cubic
Fixes a first-paint flash by applying host attribute/data/style bindings during change detection instead of after render. Prevents disabled buttons and roving-focus tabbability from flashing on client navigation; SSR unaffected.

- **Bug Fixes**
  - Host bindings: new `bindingEffect` sets the initial value during CD and keeps reactive updates after render for `attrBinding`/`dataBinding`/`styleBinding` in `ng-primitives/state`.
  - Roving focus: items register synchronously in `ng-primitives/roving-focus` so the active item is tabbable before a `focus-trap` moves focus.
  - `toggle-group`: coerce `null` to `[]` on write to keep reads safe with the earlier binding timing.

<sup>Written for commit f316882bf61ac5f8fba66623dcf4190fa2622cfe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ng-primitives/ng-primitives/pull/771?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -59,6 +59,10 @@ See `.claude/rules/` for detailed coding standards:
 
 For code review, use the `ngp-code-review` skill — it consolidates these rules with the custom workspace lint rules, test conventions, and PR checklist.
 
+### Comments
+
+Keep comments short and dense. Explain _why_ (the non-obvious reason, edge case, or constraint), never _what_ the code already says. Prefer one line; only go longer when the rationale genuinely needs it, and stay terse. Don't restate the code, narrate steps, or recap history ("previously we…"). Cut a comment if the code is self-explanatory.
+
 ## Package Management
 
 - Uses pnpm with workspace configuration
--- a/packages/ng-primitives/button/src/button/button.test.ts
+++ b/packages/ng-primitives/button/src/button/button.test.ts
@@ -1,5 +1,6 @@
 import { FocusMonitor } from '@angular/cdk/a11y';
 import { Component, Input } from '@angular/core';
+import { TestBed } from '@angular/core/testing';
 import { fireEvent, render } from '@testing-library/angular';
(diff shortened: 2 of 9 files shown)


ashley-hunter:
Closing - the root cause is https://github.com/angular/angular/issues/57528

PR #783, opened 2026-06-17
feat(menu): expose container on overlay trigger state

### LINKED ISSUE #782: `NgpMenuTrigger` state does not expose `container` (inconsistent with `NgpTooltipTrigger`)
### Feature Type

Enhancement

### Feature Description

`injectMenuTriggerState()` returns a state object that omits `container`, so the
menu container cannot be read or set programmatically via the injected state.
`NgpTooltipTrigger` exposes `container` on its state, so the same pattern works
there but not for the menu.

### Reproduction

With the tooltip this works:

```ts
private readonly tooltipTrigger = injectTooltipTriggerState();

ngOnInit(): void {
  this.tooltipTrigger().container.set(this.hostElement.nativeElement.parentElement);
}
```

The equivalent for the menu fails, because `container` is not part of the state:

```ts
private readonly menuTrigger = injectMenuTriggerState();

ngOnInit(): void {
  // this.menuTrigger().container is undefined
  this.menuTrigger().container.set(this.hostElement.nativeElement.parentElement);
}
```

### Use case

The menu content is portaled into a container, which defaults to
`document.body`. As a result, component- or segment-scoped CSS cannot target the
rendered menu; menu styling has to be declared globally to reach it.

### Proposed change

Expose `container` on the `NgpMenuTrigger` state object, matching `NgpTooltipTrigger`, so it
can be read and set via `injectMenuTriggerState()`.

ashley-hunter:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

Closes #782

`injectMenuTriggerState()` returned a state object that omitted `container`, so the menu container could not be read or set programmatically via the injected state - unlike the documented `NgpTooltipTrigger` pattern in the issue.

This PR exposes `container` on the menu trigger state as a writable, input-synced signal (via `controlled`) with a `setContainer()` setter. Direct `.set()` is routed through the setter and deprecated via `deprecatedSetter`, matching the rest of the state API.

While auditing the other overlay-based triggers, the same gap existed elsewhere, so the fix is applied consistently across all of them:

- **context-menu**: `container` was completely absent from the state (identical bug to the menu).
- **tooltip / popover / select / navigation-menu**: `container` was exposed read-only and could not be set programmatically.

Each now exposes a writable, input-synced `container` signal plus a `setContainer()` method, giving a uniform `state().setContainer(...)` API across every overlay trigger.

### Tests

Added a test per primitive that injects the state, calls `setContainer(host)` in `ngOnInit`, opens the overlay, and asserts the content portals into the custom container. Full suite: 1406 passing.

- [ ] Yes
- [x] No

---
## Summary by cubic
Expose a writable, input-synced container on all overlay trigger states and add setContainer(), enabling programmatic control of where overlays render. Fixes missing/readonly container on `menu`, `context-menu`, `tooltip`, `popover`, `select`, and `navigation-menu` to match the documented pattern (fixes #782).

- **New Features**
  - Added writable `container` signal and `setContainer()`; direct `.set()` is routed through the setter and deprecated via `deprecatedSetter`.
  - Behavior: setter takes effect on the next open and defaults to `'body'`; use via injected state (`state().setContainer(host)`).
  - Added tests verifying content portals into a custom host for each primitive.

<sup>Written for commit 3e6d210d381df1f65c73fd7685d7154b37c269a9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ng-primitives/ng-primitives/pull/783?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ng-primitives/context-menu/src/context-menu-trigger/context-menu-trigger-state.ts
+++ b/packages/ng-primitives/context-menu/src/context-menu-trigger/context-menu-trigger-state.ts
@@ -1,5 +1,13 @@
 import { FocusOrigin } from '@angular/cdk/a11y';
-import { computed, inject, Injector, signal, Signal, ViewContainerRef } from '@angular/core';
+import {
+  computed,
+  inject,
+  Injector,
+  signal,
+  Signal,
+  ViewContainerRef,
+  WritableSignal,
+} from '@angular/core';
 import { Placement } from '@floating-ui/dom';
 import { injectElementRef } from 'ng-primitives/internal';
 import { NgpMenuTriggerStateToken } from 'ng-primitives/menu';
@@ -17,6 +25,7 @@ import {
   controlled,
   createPrimitive,
   dataBinding,
+  deprecatedSetter,
   listener,
   onDestroy,
   StateInjectionOptions,
@@ -35,6 +44,12 @@ export interface NgpContextMenuTriggerState {
   readonly openOrigin: Signal<FocusOrigin>;
 
   /**
+   * The container in which the menu should be attached.
+   * @default document.body
+   */
+  readonly container: WritableSignal<HTMLElement | string | null>;
+
+  /**
    * Show the context menu.
    */
   show(origin?: FocusOrigin): void;
@@ -43,6 +58,12 @@ export interface NgpContextMenuTriggerState {
    * Hide the context menu.
    */
   hide(origin?: FocusOrigin): void;
+
+  /**
+   * Set the container in which the menu should be attached.
+   * @param container - The new container
+   */
+  setContainer(container: HTMLElement | string | null): void;
 
   /**
    * Set whether the pointer is over the menu content.
@@ -114,7 +135,7 @@ export const [
     offset: _offset = signal(2),
     flip: _flip = signal(true),
     context: _context = signal<T>(undefined as T),
-    container,
+    container: _container,
     scrollBehavior,
     shift,
   }: NgpContextMenuTriggerProps<T>) => {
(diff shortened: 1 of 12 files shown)


ashley-hunter:
Thanks @cubic-dev-ai. I looked into both findings (the same point on `setContainer` not rebinding an already-created overlay) and I'm declining the rebind for this PR - it's a pre-existing limitation rather than a regression, and out of scope here:

- `NgpOverlayConfig.container` is a plain value resolved once at attach time (`resolveContainer()` in `portal/src/overlay.ts`), so the overlay captures the container as a snapshot when it's created. `placement` is passed through as a live signal; `container` never was.
- The existing `container` **input** has the identical behaviour - changing `[ngp*Container]` after the overlay is open doesn't move already-rendered content either. This PR exposes the same semantics programmatically; it doesn't make anything worse.
- The issue this PR fixes (#782) sets the container in `ngOnInit`, before the overlay is ever created, which works correctly and is covered by tests.

The suggested destroy-and-recreate-on-change approach would tear down focus-trap state and restart entrance animations for an already-open overlay (a visible flicker), and it'd make the setter diverge from the input. A proper "live container" feature belongs at the overlay layer so the input and setter share it - that's a separate enhancement.

I've added a JSDoc note on each `setContainer()` clarifying it takes effect on next open.

PR #787, opened 2026-06-18
fix(select,combobox): validate active descendant when options change while open

### LINKED ISSUE #775: [Select/Combobox] Active option can become invalid when filtering options
### Affected Primitive

Select, Combobox

### Version

0.120.5

### Bug Description

When filtering options while the dropdown is open, the current active option can be removed from the DOM.

In that case, aria-activedescendant may point to an element that no longer exists.

This was discussed in #772, but it is not specific to that PR. The same issue seems to affect `ngpCombobox`.

A possible fix could be to call `activeDescendantManager.validate()` when the options change while the dropdown is open.

### Steps to Reproduce

1. Open a Select or Combobox.
2. Move the active option to an item.
3. Filter the list so that item is removed.
4. Check aria-activedescendant.

### Reproduction

_No response_

mvanhorn:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

Closes #775

When filtering options while a Select or Combobox dropdown is open, the currently active option can be removed from the DOM. The active-descendant manager keeps its previous active index, so `aria-activedescendant` ends up pointing at an element id that no longer exists, which breaks screen-reader announcement and keyboard navigation. As noted in the issue, this affects `ngpCombobox` as well as Select.

The `activeDescendantManager` already exposes a `validate()` method whose documented purpose is exactly "any time the item list changes, check if the active index is still valid", but it was never invoked when the options list changed while the dropdown was open.

This PR adds an open-only reactive effect in both `combobox.ts` and `select-state.ts` that observes the sorted options signal and calls `activeDescendantManager.validate()` when `open()` is true, clamping the active index back into the valid range (or onto the first enabled option) whenever options are added or removed. The `validate()` call is wrapped in `untracked()` so it does not register the manager's own internal signals as effect dependencies. Closed dropdowns are deliberately left untouched, so validating a closed list never reactivates options. The change uses only the existing public `validate()` API, so no change to `active-descendant.ts` was needed, and it is symmetric across the two components so Select and Combobox behave identically.

Tests added to `combobox.test.ts` and `select-primitive.test.ts` cover: the active option being filtered out while open (asserts `aria-activedescendant` references a still-rendered option and the stale id is gone), filtering down to zero matching options while open (asserts no stale `aria-activedescendant` and no error), and options changing while the dropdown is closed (asserts `validate()` is not triggered and the active index is untouched).

- [ ] Yes
- [x] No

---
## Summary by cubic
Fixes an a11y bug where filtering or virtual list changes while a `NgpSelect` or `NgpCombobox` is open could leave `aria-activedescendant` pointing to a removed option, breaking screen reader and keyboard navigation. Revalidates the active descendant whenever visible or backing option lists change to meet #775.

- **Bug Fixes**
  - Revalidate the active index when `sortedOptions()` or `allOptions()` change while open.
  - Use `activeDescendantManager.validate()` wrapped in `untracked()`; keep closed lists untouched.
  - Keep `aria-activedescendant` in sync, clamping to the first enabled option or clearing when empty; add tests for Select and Combobox, including virtual list scenarios.

<sup>Written for commit 698be7ae91eda8fefa963fc2a7b1dcad04ddfbbe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ng-primitives/ng-primitives/pull/787?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ng-primitives/combobox/src/combobox/combobox.test.ts
+++ b/packages/ng-primitives/combobox/src/combobox/combobox.test.ts
@@ -1,5 +1,6 @@
 import {
   Component,
+  Directive,
   TemplateRef,
   ViewContainerRef,
   forwardRef,
@@ -25,6 +26,14 @@ import { NgpComboboxInput } from '../combobox-input/combobox-input';
 import { NgpComboboxOption } from '../combobox-option/combobox-option';
 import { NgpComboboxPortal } from '../combobox-portal/combobox-portal';
 import { NgpCombobox } from './combobox';
+import { injectComboboxState } from './combobox-state';
+
+@Directive({
+  selector: '[captureComboboxState]',
+})
+class CaptureComboboxStateDirective {
+  readonly state = injectComboboxState();
+}
 
 @Component({
   imports: [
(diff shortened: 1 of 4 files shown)


ashley-hunter (approved):
Nice work, thank you very much for the contribution!

ashley-hunter:
@allcontributors add @mvanhorn for code

mvanhorn:
Thanks @ashley-hunter for taking this in. Validating the active descendant in select and combobox tightens up the a11y behavior.

ashley-hunter:
Absolutely! Thanks very much for the contribution! Nice improvement for sure!

mvanhorn:
Thank you @ashley-hunter. Validating the active descendant when options change is the kind of a11y detail that's easy to overlook.

PR #808, opened 2026-07-04
fix(menu): focus non-native menu items

### LINKED ISSUE #800: `ngpMenuItem` does not receive focus when used on non-focusable elements
### Affected Primitive

Menu

### Version

0.122.0

### Bug Description

When using `ngpMenu`, the first menu item is not focused when the menu opens if `ngpMenuItem` is applied to a non-focusable element, such as a `div`.

I am raising this case because in my use case, each menu item displays additional actions on hover. Because of that, I intentionally do not want to use a native `button` for the menu item itself, as it would lead to nested `button` elements when rendering action buttons inside the item.

For example, this does not focus any item when navigating the menu:

```html
<button [ngpMenuTrigger]="menu" ngpButton>Open Menu</button>
<ng-template #menu>
  <div ngpMenu>
    <div ngpMenuItem>Item 1</div>
    <div ngpMenuItem>Item 2</div>
    <div ngpMenuItem>Item 3</div>
  </div>
</ng-template>
```

However, the same menu works correctly when `ngpMenuItem` is used on native focusable elements such as `button`:

```html
    <button ngpMenuItem>Item 1</button>
```

A workaround is to manually add `tabindex="0"` to the first item:

```html
<div ngpMenuItem tabindex="0">Item 1</div>
```

### Steps to Reproduce

1. Create a menu using `ngpMenu`.
2. Add menu items using `ngpMenuItem` on `div` elements.
3. Open the menu.
4. Try to navigate the menu with the keyboard.
5. Notice that no menu item receives focus.
6. Replace the `div` elements with `button` elements.
7. Open the menu again.
8. Notice that keyboard focus works correctly.

### Reproduction

https://stackblitz.com/edit/xuccldeu

xonaib:
## Summary

Fixes focus for menu items rendered on non-native focusable elements, like `div`.

These items already receive the right `tabindex`, but focus can stay on the menu container when the menu opens. This updates the menu to hand focus to the first roving-focus item after render, while avoiding stealing focus during submenu interactions.

Fixes #800

## Testing

- `.\node_modules\.bin\vitest.CMD --config packages\ng-primitives\vite.config.ts packages\ng-primitives\menu\src\menu\menu.test.ts --run`
- `.\node_modules\.bin\nx.CMD build ng-primitives --excludeTaskDependencies`

---
## Summary by cubic
Fixes menu focus so `ngpMenuItem` on non-native elements (like `div`) receives focus on open, not the menu container. Prevents focus stealing during submenu interactions.

- **Bug Fixes**
  - After render, if the menu container is focused, return focus to the active roving-focus item (uses `afterNextRender` + `requestAnimationFrame`).
  - Injects roving-focus group state to activate the first item based on `openOrigin`.
  - Adds a test ensuring the first `div`-based item is focused on open and arrow navigation works.

<sup>Written for commit 96f62ff02b5d7576c08b53d1ff2f481251b024eb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ng-primitives/ng-primitives/pull/808?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ng-primitives/menu/src/menu/menu-state.ts
+++ b/packages/ng-primitives/menu/src/menu/menu-state.ts
@@ -1,10 +1,12 @@
 import { FocusOrigin } from '@angular/cdk/a11y';
 import { Directionality } from '@angular/cdk/bidi';
-import { computed, inject } from '@angular/core';
+import { afterNextRender, computed, inject, Injector } from '@angular/core';
 import { ngpFocusTrap } from 'ng-primitives/focus-trap';
 import { injectElementRef } from 'ng-primitives/internal';
 import { injectOverlay } from 'ng-primitives/portal';
+import { injectRovingFocusGroupState } from 'ng-primitives/roving-focus';
 import { attrBinding, createPrimitive, listener, styleBinding } from 'ng-primitives/state';
+import { injectDisposables } from 'ng-primitives/utils';
 import { Subject } from 'rxjs';
 import { injectMenuTriggerState } from '../menu-trigger/menu-trigger-state';
 
@@ -29,8 +31,11 @@ export const [NgpMenuStateToken, ngpMenu, injectMenuState, provideMenuState] = c
     const element = injectElementRef();
     const overlay = injectOverlay();
     const directionality = inject(Directionality, { optional: true });
+    const injector = inject(Injector);
+    const disposables = injectDisposables();
     const menuTrigger = injectMenuTriggerState();
     const parentMenu = injectMenuState({ optional: true, skipSelf: true });
+    const rovingFocusGroup = injectRovingFocusGroupState();
 
     // Always trap focus in menus per WAI-ARIA guidelines (Tab should not navigate within menus)
     // Pass the open origin so focus trap uses the correct origin for :focus-visible styling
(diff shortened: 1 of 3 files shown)


ashley-hunter:
Thanks very much for the contribution!

While this resolves the issue, I think this actually exposes that there is an underlying race condition.

I think it might be better to fix the underlying issue here rather than work around the timing. The afterNextRender + requestAnimationFrame gets the right result, but it’s dodging a phase-ordering race: roving-focus applies each item’s tabindex in a later render phase than the focus trap’s initial-focus scan, and that scan only finds elements that are already tabbable. A <button> qualifies no matter what, a <div> only once tabindex="0" has landed, so focus falls back to the container. The activeElement === container guard is really there to catch that fallback.

The proper fix is to sort out who owns initial focus. Right now the focus trap both traps and places focus, and the menu leans on that. Instead the trap should only trap, and the menu should place initial focus itself:

	•	Add an autoFocus opt-out to the focus trap (default on, so nothing else changes). When off it still traps, it just doesn’t place initial focus.
	•	Have the menu pass autoFocus: false and call rovingFocusGroup.activateFirst(openOrigin()) itself, in the read after-render phase so it runs after the items have applied their tabindex.

That removes the requestAnimationFrame and the active-element guard entirely, and div and button items take the same deterministic path.

Hope that makes sense! Thanks

PR #841, opened 2026-07-21
fix(schematics): generate templates that compile

### LINKED ISSUE #840: `ng g ng-primitives:primitive` generates code that does not compile for popover, tabs, tooltip and input-otp
### Affected Primitive

Schematics — popover, tabs, tooltip, input-otp

### Version

0.127.0

### Bug Description

Generating any primitive that is split across two files produces a sibling import that
resolves to nothing. `popover-trigger.component.ts` contains
`import { Popover } from './popover'`, but the file it is generated next to is called
`popover.component.ts` and exports `PopoverComponent`. `tabs` and `tooltip` are the same.
It only resolves if you pass both `--file-suffix ""` and `--component-suffix ""`, which is
neither the default nor what the docs show.

Separately, `input-otp` generates a file that does not parse at all: the
`ChangeDetectionStrategy` import rewrite leaves a double comma on a wrapped import list.
That fires by default, since `OnPush` is the default.

Nothing catches either today — the schematic tests assert on strings rather than compiling
anything, and Prettier skips `.template` files, so the generated output has never been
compiled by anything.

### Steps to Reproduce

In any Angular workspace with `ng-primitives` installed:

```bash
ng g ng-primitives:primitive popover --path src/app/popover
```

Two files are created. `src/app/popover/popover-trigger.component.ts` starts with:

```ts
import { Popover } from './popover';
```

There is no `./popover` — the sibling is `popover.component.ts` — and it exports
`PopoverComponent`, not `Popover`. Two errors: TS2307 on the module, then the symbol.

Same for `ng g ng-primitives:primitive tabs` and `ng g ng-primitives:primitive tooltip`.

For the parse failure:

```bash
ng g ng-primitives:primitive input-otp --path src/app/input-otp
```

The generated file opens with `import { booleanAttribute, ..., numberAttribute,,
ChangeDetectionStrategy } from '@angular/core';` — TS1003.

### Reproduction

_No response_

pauldvlp:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

Closes #840 

`ng g ng-primitives:primitive popover`, `tabs` or `tooltip` generates code that does not
compile, and `input-otp` generates code that does not parse.

### What was broken

The templates generator's guard for relative imports read the string literal with
`getText()`, which keeps the surrounding quotes, so `"'./popover'"` never looked relative
and both import transforms were permanent no-ops. The generated part therefore imported
`{ Popover } from './popover'` next to a file named `popover.component.ts` exporting
`PopoverComponent`. It only resolved if you passed both `--file-suffix ""` and
`--component-suffix ""`.

Rewriting the specifier alone was not enough. Imported symbols were renamed only inside
the import declaration, so `tabs.ts`'s `contentChildren(Tab)` kept the unsuffixed name.
Whether a symbol takes the suffix depends on the sibling declaring it as a component — a
file cannot know that from its own source — so the generator now collects the component
classes of the whole primitive before rewriting anything. `@Directive` parts such as
`PopoverTrigger` deliberately keep their name.

Rewrites are now collected against a single parse and applied right to left. The four
mutation loops used to splice while iterating nodes from a stale parse, and were correct
only because every file happens to have one class, one selector and at most one relative
import.

Separately, `processChangeDetection`'s import rewrite left a double comma on a wrapped
import list, so `input-otp` generated a file that did not parse. That fired by default,
since `OnPush` is the default.

### Why nothing caught it

Prettier reports `inferredParser: null` for `.template`, so `nx format:check` skips them,
and the schematic tests assert on strings rather than compiling anything. The 46 template
files had never been compiled by anything.

This PR adds a gate that generates every primitive in the schema and type-checks it with
`performCompilation` from `@angular/compiler-cli` under `strictTemplates` (~5s, on the
existing `test-node` target). It is what surfaced all four bugs. Parsing is asserted
separately, because `performCompilation` gives up on the first syntactic error and would
otherwise hide everything behind it.

The gate runs unconditionally in CI rather than through `nx affected`, alongside a
`git diff --exit-code` sync check:

```
$ npx nx show projects --affected --files=apps/components/.../switch/switch.ts
["components"]
$ npx nx show projects --affected --files=packages/tools/.../templates.ts
["tools"]
```

`ng-primitives` is never affected by a change to either of the two inputs its templates are
generated from, so an `affected`-driven check would never fire on the PRs that matter.

### The regenerated templates

Three change because of the fix. The other nine were stale for the same reason: nothing
regenerates them on a pull request. No consumer was affected — `nx-release-publish` depends
on `build`, which regenerates before packing, so npm has always shipped a freshly generated
copy. The second commit only makes the committed files match what is published.

### Tests and docs

`compiles.node.test.ts` is new; `index.node.test.ts` gains eight cases covering the three
multi-file primitives across suffix combinations, the dasherised file suffix, and the usage
sites. Six of them fail against the previous templates.

No documentation change: nothing user-facing is documented about how a generated part
imports its sibling.

### Not included

Found while working on this, each worth its own issue rather than widening this PR: the
`build:schematics` asset glob names `migrations.json` while the file is `migration.json`;
the selector rewrite is a literal `'app-'` replace, so `--prefix` is silently ignored for
the two camelCase attribute selectors; and four primitives have templates but are missing
from `schema.json`'s enum, two of which publish an `ng g` command in their docs page that
fails validation today.

- [ ] Yes
- [x] No

Generated output changes for `popover`, `tabs`, `tooltip` and `input-otp`, but only from
code that did not compile to code that does. Existing generated components in consumer
projects are untouched.

---
## Summary by cubic
Fixes the schematics generator so all generated primitives compile and adds a CI gate to prevent regressions. Regenerates templates with ControlValueAccessor and disabled-state fixes.

- **Bug Fixes**
  - Cross-file references resolve for `popover`, `tabs`, and `tooltip`: rewrite relative import paths with an optional dasherized `<%= fileSuffix %>` and apply `<%= componentSuffix %>` to component classes and all usage sites.
  - Fixed `ChangeDetectionStrategy` import rewrite to avoid a trailing double comma that broke `input-otp`.
  - CVA and UX fixes:
    - Prevent re-emitting on `writeValue` in `checkbox`, `date-picker`, `slider`, `range-slider`, `switch`, and `pagination`.
    - Propagate form disabled state (`listbox`, `pagination`) and mark touch on `listbox` focusout.
    - `combobox`: accept `undefined` in `onChange` and clear the form value when the filter is cleared.
    - Corrected `toggle-group` input mappings.

- **Refactors**
  - Generator collects component classes across a primitive and applies all AST edits in one pass; skips example-only `index.page.ts` files.
  - Added a `test-node` gate that generates every primitive and type-checks with `@angular/compiler-cli` (`strictTemplates`); runs unconditionally in CI with a sync check (`nx run ng-primitives:build:templates` + `git diff --exit-code`).
  - Made `build:schematics:test` depend on `build:templates`.

<sup>Written for commit 98c48346c2dcb9ed11484357152b9a333d972a6e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ng-primitives/ng-primitives/pull/841?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved form control synchronisation across generated primitives by suppressing duplicate change emissions during model-to-view updates (checkbox, date-picker, combobox, range-slider, slider, switch, listbox, pagination).
  - Disabled-state handling now propagates correctly where supported.
  - Fixed generated typing and imports for suffixed primitives (tabs, popover-trigger, tooltip-trigger), and corrected toggle-group input mappings.
- **Quality Improvements**
  - Added stronger validation to ensure schematic templates stay in sync and that all generated primitives compile/type-check successfully.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,18 @@ jobs:
       - uses: nrwl/nx-set-shas@v4
 
       - run: pnpm exec nx format:check
+
+      # Runs unconditionally rather than through `nx affected`: a change to
+      # apps/components or to packages/tools never marks ng-primitives as affected, so
+      # these checks would otherwise never fire on the PRs that change what the templates
+      # are generated from.
+      - name: Schematic templates are in sync and compile
+        run: |
+          pnpm exec nx run ng-primitives:build:templates
+          git diff --exit-code -- packages/ng-primitives/schematics/ng-generate/templates \
+            || { echo "::error::Schematic templates are stale. Run: nx run ng-primitives:build:templates"; exit 1; }
+          pnpm exec nx run ng-primitives:test-node
+
       - run: pnpm exec nx affected -t lint,build,test
         env:
           NODE_OPTIONS: '--max-old-space-size=8192' # Increase memory to 8GB
--- a/packages/ng-primitives/project.json
+++ b/packages/ng-primitives/project.json
@@ -65,6 +65,7 @@
     },
     "build:schematics:test": {
       "executor": "@nx/js:tsc",
+      "dependsOn": ["build:templates"],
       "outputs": ["{options.outputPath}"],
       "options": {
         "clean": true,
--- a/packages/ng-primitives/schematics/ng-generate/compiles.node.test.ts
+++ b/packages/ng-primitives/schematics/ng-generate/compiles.node.test.ts
@@ -1,0 +1,164 @@
+import { SchematicTestRunner, UnitTestTree } from '@angular-devkit/schematics/testing';
+import { performCompilation, readConfiguration } from '@angular/compiler-cli';
+import { Schema as ApplicationOptions } from '@schematics/angular/application/schema';
+import { Schema as WorkspaceOptions } from '@schematics/angular/workspace/schema';
+import { mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
+import { dirname, join, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import ts from 'typescript';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+
+/**
+ * Every primitive the schematic can generate must produce code that actually compiles.
+ *
+ * Nothing else in the repo checks this: the templates are generated build output that
+ * Prettier skips (`inferredParser: null` for `.template`), and the schematic's other
+ * tests assert on strings rather than compiling anything.
+ */
+describe('generated primitives compile', () => {
+  const workspaceRoot = resolve(fileURLToPath(new URL('../../../../', import.meta.url)));
+  const schematicRunner = new SchematicTestRunner(
+    'ng-primitives',
+    resolve(workspaceRoot, 'dist/ng-primitives-schematics-test/collection.json'),
+  );
+
+  const workspaceOptions: WorkspaceOptions = {
+    name: 'workspace',
+    newProjectRoot: 'projects',
+    version: '6.0.0',
+  };
+
+  const appOptions: ApplicationOptions = {
+    name: 'bar',
+    inlineStyle: false,
+    inlineTemplate: false,
+    routing: false,
+    skipTests: false,
+    skipPackageJson: false,
+  };
+
+  // Read at runtime rather than importing, so the gate covers the schema that ships.
+  const schemaPath = resolve(fileURLToPath(new URL('./schema.json', import.meta.url)));
+  const primitives: string[] = JSON.parse(readFileSync(schemaPath, 'utf-8')).properties.primitive
+    .enum;
+
+  /**
+   * The generated files have to live inside the repo: they import `@angular/core`,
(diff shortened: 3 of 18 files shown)


ashley-hunter (changes requested):
Looks great! Thank you! Looks like it is failing on CI - which is great it means it's working 😂 just because some listbox changes have since been merged.

pauldvlp:
Excellent! 😂 Now everything seems to be working! I'm glad I did the right thing!

ashley-hunter (approved):
Nice improvement 😊 thank you!

PR #838, opened 2026-07-21
feat(search): migrate search state

pauldvlp:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/ng-primitives/ng-primitives/blob/main/CONTRIBUTING.md#-commit-message-guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes
- [ ] Other... Please describe:

## Issue

Closes #837

## What does this PR implement/fix?

Migrates the search primitive from the legacy `createStateToken`/`createStateProvider`/`createStateInjector`/`createState` quadruple to the current `createPrimitive` state pattern, in line with the recent meter (#830, #832) and `input-otp` (#834) migrations.

### What moved

- **`search`**: the input and value signals, the empty computed, the data-empty binding and the Escape-to-clear handler all move into a `createPrimitive` factory. The directive keeps a public `empty` signal and `clear()` that delegate to the state; input and value stay private, and only `empty`, `clear` and `registerInput` are returned.
- **`search-clear`**: `new -state.ts` factory owns the `tabindex` and `data-empty` bindings and the click listener; the directive is a thin shell.
- **`registerInput`** passes the search's own injector to `listener()`, so the input-event subscription stays tied to the search lifecycle rather than the caller's, matching the previous `safeTakeUntilDestroyed(this.destroyRef)` behaviour.
- The barrel now exports the factories, injectors, providers and state types for both parts, in line with the other migrated primitives.

### Tests and docs

No new tests: behaviour is unchanged, and the existing suite already covers what moved - `data-empty` tracking, Escape-to-clear, the clear button, the disabled guard, and the directive's `empty()`/`clear()` API. No docs change either, since there is no consumer-facing API change.

Full ng-primitives suite green (2246/2246), plus lint (0 errors) and AOT build.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

---
## Summary by cubic
Migrated the `search` primitive to the `createPrimitive` pattern, moving host bindings and listeners into state factories and leaving directives as thin shells. No behavior or public API changes.

- **Refactors**
[...]

PR #834, opened 2026-07-20
feat(input-otp): migrate input-otp state

pauldvlp:
## PR Checklist

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added
- [x] Docs have been added / updated — internal state-pattern migration with no consumer-facing API change, so no docs update is needed

## PR Type

- [x] Refactoring (no functional changes, no api changes)

## Issue

Closes #833

## What does this PR implement/fix?

Migrates the input-otp primitive from the legacy createStateToken/createStateProvider/createStateInjector/createState quadruple to the current createPrimitive state pattern, in line with the recent meter migration (#830, #832).

### What moved
- input-otp: root directive logic now lives in a createPrimitive factory. value is controlled (mutated as the user types); the other inputs pass through read-only. valueChange/complete emit via onValueChange/onComplete callbacks. inputElement and the slot list stay private — only maxLength and the inter-part methods are returned.
- input-otp-input / input-otp-slot: new -state.ts factories own their host bindings and listeners; the directives are thin shells.
- Slots register via registerSlot/unregisterSlot + onDestroy (as select/dialog/roving-focus do), so maxLength and slot indices stay accurate as slots are added/removed.

### Tests
Adds a regression test covering dynamic slot removal. Full ng-primitives suite green (2246/2246), plus lint (0 errors) and AOT build. No consumer-facing API changes.

## Does this PR introduce a breaking change?

- [x] No

---
## Summary by cubic
Migrates `input-otp` to the `createPrimitive` state pattern. No API changes; adds a regression test for dynamic slot removal (closes #833).

- **Refactors**
  - Root state now provided by `ngpInputOtp`; `value` is controlled, other inputs are read-only; emits via `onValueChange` and `onComplete`.
  - `input-otp-input` and `input-otp-slot` move host bindings and listeners into new `-state.ts` factories; directives are thin shells.
  - Slots register/unregister with `onDestroy`, keeping `maxLength` and indices accurate as slots are added/removed.
  - Hidden input composes `ngpVisuallyHidden` inside its state instead of via a host directive.

<sup>Written for commit eefced007a67c48f62cb461bdd4f603af47a5d95. Summary will update on new commits.</sup>

[...]

PR #759, opened 2026-05-21
feat(combobox): migrate combobox state

MGREMY:
## PR Checklist

~- [x] Fix tests~

Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/ng-primitives/ng-primitives/blob/main/CONTRIBUTING.md#-commit-message-guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes
- [ ] Other... Please describe:

## Issue

Closes #560 

## What does this PR implement/fix?

Migrate state management

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

---
## Summary by cubic
Migrated Combobox to the new primitive state pattern with dedicated state modules (`button`, `input`, `option`, `dropdown`, `portal`). Improves ARIA/data bindings, keyboard/selection behavior, overlay lifecycle, and cleans up exports and tests.

- **Refactors**
  - Introduced state primitives via `createPrimitive`/`controlled` for `button`/`input`/`option`/`dropdown`/`portal`; directives are thin wrappers with providers.
  - Centralized open/close, selection, active-descendant, and focus in `ngpCombobox`; added host keyboard support when no input is present; wired overlay/portal for positioning and attach/detach lifecycle.
  - Added virtual scrolling hooks via option `index` and `scrollToOption`.
  - Strengthened ARIA roles/attributes and `data-*` bindings; applied control-status data attrs on the host via `ngpFormControl`.
  - Cleaned exports: re-export state tokens/hooks/types and `NgpComboboxPlacement` from `combobox/combobox-state` at the package root; updated `NgpComboboxConfig` import.
  - Updated utils/tests: `areAllOptionsSelected` now uses option state; stabilized async tests with `waitFor`.
  - Fixes: `valueChange` emits on programmatic set; removed nested keydown listener; handled `undefined` in `selectOption`; ensured overlay detaches on destroy; used the value signal in state; handled `resetOnClose` on same-value emit; added `SetterOptions` to `setValue`; added missing bindings/providers.

- **Migration**
[...]

PR #849, opened 2026-07-23
feat(date-picker): migrate date-picker state

### LINKED ISSUE #562: Date Picker State Pattern Migration
Migrate date picker to new state pattern

pauldvlp:
## PR Checklist

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added
- [x] Docs have been added / updated — internal state-pattern migration with no consumer-facing API change, so no docs update is needed

## PR Type

## Issue

Closes #562

Migrates the `date-picker` primitive (`date-picker` and `date-range-picker`, plus their 8 parts) from the legacy createStateToken/createStateProvider/createStateInjector/createState quadruple to the current createPrimitive state pattern, in line with the recent listbox (#835), search (#838), input-otp (#834) and ai (#846) migrations. With combobox (#759) in progress, this removes the last remaining use of the legacy state helpers.

### What moved

- **date-picker / date-range-picker (roots):** both rewritten with `createPrimitive`. They satisfy a shared `NgpDateControllerState<T>` interface that every part reads through `injectDateControllerState`, which resolves to whichever container (single or range) is present. `min`/`max`/`dateDisabled`/`firstDayOfWeek` pass through read-only; `date`/`startDate`/`endDate`/`focusedDate`/`disabled` are `controlled` writable signals; `dateChange`/`focusedDateChange` are emitters exposed as observables, with the directive outputs bridged through `on*Change` callbacks.
- **date buttons** register with the container via `registerButton`/`unregisterButton` (the backing list stays private to the factory), the way roving-focus registers its items — a small state object exposing `focus()`/`selected`/`disabled` rather than the directive instance, deregistering with `onDestroy`. The keyboard-focus `afterNextRender` loop moves into the factory.
- **label ↔ grid:** the label registers its id via `setLabel`/`removeLabel` (single-value reseed guard) and the grid reflects it on `aria-labelledby` through `labelledBy`, replacing the root `contentChild` + `state().label()?.id()`.
- **date-button / next-month / previous-month / grid / label / cell:** host bindings and listeners move into `-state.ts` factories; the month buttons compose `ngpButton({ disabled })`. The date button's `@HostListener('keydown.arrowLeft')` (and the other pseudo-events) become a single `keydown` listener switching on `event.key`, since `listener()` registers raw DOM events. A modifier guard preserves Angular's pseudo-event semantics — combos like Alt+ArrowLeft (browser back) and Shift+PageUp/PageDown pass through instead of being handled and `preventDefault`-ed.
- **cell-render / row-render** are unchanged: they have no host bindings and only consume the controller state.
- `disabled` on the injected state is exposed through `deprecatedSetter` with a paired `setDisabled`; the reusable date-picker CVA and its generated schematic template switch to `state().setDisabled(...)`.

### Tests

Adds a regression test that a navigation key held with a modifier (Alt+ArrowLeft) neither moves focus nor calls `preventDefault`. Full date-picker suite green (80/80), plus lint and AOT build.

- [ ] Yes
- [x] No

---
## Summary by cubic
Migrates `date-picker` and `date-range-picker` to `createPrimitive` with a shared `NgpDateControllerState`, and adopts `controlledState` with new default-value inputs for controlled/uncontrolled latching. No breaking changes; closes #562.

- **Refactors**
  - Rebuilt with `ngpDatePicker`/`ngpDateRangePicker`; parts read a shared controller via `injectDateControllerState`.
  - Adopted `controlledState` for two-way values; added `ngpDatePickerDefaultDate`/`ngpDatePickerDefaultFocusedDate` and `ngpDateRangePickerDefaultStartDate`/`DefaultEndDate`/`DefaultFocusedDate`, plus `setDefault*` setters and latching behavior.
  - Moved host bindings/listeners into part `*-state.ts`; date buttons get `provideDatePickerDateButtonState()` and a unified `keydown` handler; month buttons compose `ngpButton`.
  - Expose `dateChange`/`startDateChange`/`endDateChange` as emitters; removed unused `focusedDateChange` emitter (focused date uses the callback only).
  - Added `setDisabled(...)` with a `deprecatedSetter` bridge; CVA and schematics now call `state().setDisabled(...)`.
  - Label registers its id and the grid reflects it via `aria-labelledby`; `index.ts` exports new state tokens/factories and types; range `select(...)` honors `{ emit: false }`.

- **Bug Fixes**
  - Keyboard navigation ignores modified keys (e.g., Alt+ArrowLeft) and doesn’t `preventDefault`; regression test added.

<sup>Written for commit 8058ed17817e2d1dfd9f717b2cea516efcb9dc9d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ng-primitives/ng-primitives/pull/849?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added uncontrolled default inputs for the date picker and date range picker (including default dates and default focused date).
  * Improved keyboard navigation across the calendar, including RTL-aware arrow key movement and month-to-month transitions.
* **Accessibility**
  * Enhanced ARIA/data semantics consistency across the grid, labels, and date cells.
  * Previous/next month controls now reliably clamp day-of-month when moving to shorter months.
* **Bug Fixes**
  * Disabled-state updates now apply consistently, and modifier-key keyboard behaviour no longer changes roving focus.
* **Tests**
  * Added an Alt+ArrowLeft regression test and expanded controlled/uncontrolled selection coverage.


packages/ng-primitives/date-picker/src/date-picker/date-picker-state.ts
```diff
@@ -1,32 +1,348 @@
...
+    onDateChange,
+    onFocusedDateChange,
+  }: NgpDatePickerProps<T>) => {
+    const elementRef = injectElementRef();
+    const dateAdapter = injectDateAdapter<T>();
+    const injector = inject(Injector);
+
+    const disabled = controlled(_disabled);
+    const focusedDate = controlled(_focusedDate, dateAdapter.now());
+    const dateValue = controlled(_date);
```
ashley-hunter: I believe the correct choice here would be to use controlledState rather than controlled here - which would also require no default value to be set.

The same would also need to be done for start and end on the range picker. 

Arguably focusedDate should also use it too.

ashley-hunter (changes requested):
Thank you! Looks great! Just noticed one small thing we should change.

pauldvlp:
Done @ashley-hunter! Excellent approach indeed, Thank you! ☺️

ashley-hunter (approved):
Thank you, great work

PR #871, opened 2026-07-27
fix(ai): apply the thread viewport threshold and gate auto-scroll on it

### LINKED ISSUE #870: Thread viewport never applies ngpThreadViewportThreshold
### Affected Primitive

AI Assistant (`NgpThreadViewport`)

### Version

0.127.0

### Bug Description

`ngpThreadViewportThreshold` is documented as the distance from the bottom within which the thread is treated as being "at the bottom", and its doc comment says the value decides "whether automatic scrolling to the bottom should occur, for example when new content is added or the container is resized". Neither half of that holds today:

1. The input is never read. `onScroll` computes `scrollHeight - scrollTop <= clientHeight`, i.e. exactly at the bottom, so the tolerance is always 0 px. Sub-pixel scroll offsets (fractional device pixel ratios, zoom) can make this false even when the thread looks pinned to the bottom.
2. The resulting `isAtBottom` flag is only consulted in the resize handler. The "new content arrived" path does not consult it at all: `NgpThreadMessage`'s mutation observer calls `thread.scrollToBottom('smooth')` whenever the last message changes, so a streaming reply drags the viewport back down while the user is reading earlier messages.

So the public input is effectively a no-op, and the auto-scroll behaviour it is meant to govern ignores the user's scroll position.

Suggested fix:

- apply the threshold: `scrollHeight - scrollTop - clientHeight <= threshold()`, and pass the input through to the viewport state factory (it currently stops at the directive);
- gate the "new content" scroll on the at-bottom state, while keeping the explicit scroll on submit unconditional — the user who just sent a prompt should always be taken to the bottom. That needs the two call sites distinguished, e.g. an `autoScrollToBottom()` used by the message path alongside the existing `scrollToBottom()`.

I'd rather agree the shape of the second part before implementing it, since it changes observable behaviour.

### Steps to Reproduce

1. Open the AI Assistant example in the docs.
2. Send a message and, while the assistant reply is streaming, scroll up to read an earlier message.
3. The viewport is pulled back to the bottom on every chunk.
4. Setting `ngpThreadViewportThreshold` to any value changes nothing.

### Reproduction

Docs example: https://angularprimitives.com/primitives/ai-assistant

pauldvlp:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

Closes #870

`ngpThreadViewportThreshold` is documented as the distance from the bottom within which the thread counts as being "at the bottom", governing "whether automatic scrolling to the bottom should occur". Neither half held.

**The input never reached the state.** `NgpThreadViewport` declared it, but only `autoScroll` was passed to `ngpThreadViewport(...)`, and `onScroll` compared `scrollHeight - scrollTop <= clientHeight` — a fixed 0px tolerance. Setting the input changed nothing, and sub-pixel scroll offsets (fractional device pixel ratios, zoom) could read as "not at the bottom" while the thread looked pinned to it.

**The at-bottom result was only consulted on resize.** The new-content path ignored it: `NgpThreadMessage`'s mutation observer called `thread.scrollToBottom('smooth')` whenever the last message changed, so a streaming reply pulled the viewport back down while the user was reading earlier messages.

This PR:

- adds `threshold` to `NgpThreadViewportProps`, passes the input through from the directive, and measures `scrollHeight - scrollTop - clientHeight <= threshold()`;
- adds `autoScrollToBottom(behavior)`, which scrolls only when the viewport is already at the bottom, and routes the thread-message path through it. The resize handler now shares it instead of repeating the check inline.

The explicit scroll on submit still goes through `scrollToBottom` and stays unconditional — someone who just sent a prompt should always be taken to the bottom.

Four tests cover it: streaming does not move a viewport the user scrolled up from, still follows one that is at the bottom, treats a position inside the threshold as at the bottom, and honours a custom threshold. The first and last fail on `next`.

No documentation change: the input's doc comment already describes this behaviour and is the source for the generated API reference — it was accurate, the implementation was not.

- [ ] Yes
- [x] No

No public API changes — the input already existed, it just did nothing. Runtime behaviour does change, in the direction the input documents: a thread the user has scrolled up from no longer jumps to the bottom as content streams in, and `ngpThreadViewportThreshold` now takes effect where it previously had none.

---
## Summary by cubic
Fixes chat auto-scroll so it only follows streaming replies when the viewport is already at the bottom. The `ngpThreadViewportThreshold` input now works, updates live, and prevents jumps while reading.

- **Bug Fixes**
  - Pass `ngpThreadViewportThreshold` to `ngpThreadViewport` state and use it in the at-bottom check (`scrollHeight - scrollTop - clientHeight <= threshold()`), defaulting to 70px.
  - Add `scrollToBottomIfNeeded(behavior)` and use it for streaming and resize; only scroll when already at the bottom. Submit still uses `scrollToBottom` unconditionally.
  - Recompute the at-bottom state when the threshold changes to avoid stale values; tests cover scrolled-up, at-bottom, threshold tolerance, custom threshold, and live threshold changes.

<sup>Written for commit 33b3b49ba1f4c804578c4ecf2d0878db870c7462. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ng-primitives/ng-primitives/pull/871?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added configurable “at bottom” threshold for thread viewport auto-scrolling, including the ability to disable thresholding.
* **Bug Fixes**
  * Improved streaming updates so the view follows new messages only when the user is at, or within the threshold of, the bottom (otherwise it won’t interrupt older content).
  * Ensured resize and threshold changes re-check “at bottom” and only scroll when appropriate.
* **Tests**
  * Added coverage for scrolling threshold behaviour during streaming, including edge cases like custom thresholds.

### DIFF
--- a/packages/ng-primitives/ai/src/thread-viewport/thread-viewport-state.ts
+++ b/packages/ng-primitives/ai/src/thread-viewport/thread-viewport-state.ts
@@ -10,6 +10,12 @@ export interface NgpThreadViewportState {
    * Scroll the viewport to the bottom.
    */
   scrollToBottom(behavior: ScrollBehavior): void;
+  /**
+   * @internal
+   * Scroll the viewport to the bottom, but only if it is already at the bottom, so content
+   * arriving while the user is reading further up does not pull them away from it.
+   */
+  autoScrollToBottom(behavior: ScrollBehavior): void;
 }
 
 export interface NgpThreadViewportProps {
@@ -17,6 +23,10 @@ export interface NgpThreadViewportProps {
    * Whether the thread should automatically scroll to the bottom when new content is added.
    */
   readonly autoScroll?: Signal<boolean>;
+  /**
+   * The distance in pixels from the bottom that is still considered "at the bottom".
+   */
+  readonly threshold?: Signal<number>;
 }
 
 export const [
@@ -26,7 +36,10 @@ export const [
   provideThreadViewportState,
 ] = createPrimitive(
   'NgpThreadViewport',
-  ({ autoScroll = signal(true) }: NgpThreadViewportProps): NgpThreadViewportState => {
+  ({
+    autoScroll = signal(true),
+    threshold = signal(70),
+  }: NgpThreadViewportProps): NgpThreadViewportState => {
     const element = injectElementRef<HTMLElement>();
     const thread = injectThreadState();
 
@@ -47,9 +60,18 @@ export const [
       });
     }
 
+    function autoScrollToBottom(behavior: ScrollBehavior): void {
+      // the user has scrolled away from the bottom, so leave them where they are
+      if (!isAtBottom) {
+        return;
+      }
+
+      scrollToBottom(behavior);
+    }
+
     function onScroll(): void {
       const { scrollHeight, scrollTop, clientHeight } = element.nativeElement;
-      const atBottom = scrollHeight - scrollTop <= clientHeight;
+      const atBottom = scrollHeight - scrollTop - clientHeight <= threshold();
 
       if (atBottom || lastScrollTop >= scrollTop) {
         isAtBottom = atBottom;
@@ -64,13 +86,11 @@ export const [
     fromResizeEvent(element.nativeElement)
       .pipe(safeTakeUntilDestroyed())
       .subscribe(() => {
-        if (isAtBottom) {
-          scrollToBottom('instant');
-        }
+        autoScrollToBottom('instant');
         onScroll();
       });
 
-    const state = { scrollToBottom } satisfies NgpThreadViewportState;
+    const state = { scrollToBottom, autoScrollToBottom } satisfies NgpThreadViewportState;
 
     thread().setViewport(state);
     onDestroy(() => thread().removeViewport(state));
--- a/packages/ng-primitives/ai/src/thread-viewport/thread-viewport-state.ts
+++ b/packages/ng-primitives/ai/src/thread-viewport/thread-viewport-state.ts
@@ -1,5 +1,5 @@
 import { signal, Signal } from '@angular/core';
-import { fromResizeEvent, injectElementRef } from 'ng-primitives/internal';
+import { explicitEffect, fromResizeEvent, injectElementRef } from 'ng-primitives/internal';
 import { createPrimitive, listener, onDestroy } from 'ng-primitives/state';
 import { safeTakeUntilDestroyed } from 'ng-primitives/utils';
 import { injectThreadState } from '../thread/thread-state';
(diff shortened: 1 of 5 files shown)


packages/ng-primitives/ai/src/thread/thread-state.ts
```diff
@@ -64,6 +70,10 @@ export const [NgpThreadStateToken, ngpThread, injectThreadState, provideThreadSt
       viewport()?.scrollToBottom(behavior);
     }
 
+    function autoScrollToBottom(behavior: ScrollBehavior): void {
```
ashley-hunter: The DOM uses scrollIntoViewIfNeeded, I wonder should we align, eg scrollToBottomIfNeeded?

pauldvlp: Good call, renamed to scrollToBottomIfNeeded.
It also clears up an overlap I'd missed: autoScrollToBottom read as "the scroll governed by ngpThreadViewportAutoScroll", but that input gates plain scrollToBottom too, so the name pointed at the wrong condition. IfNeeded names the actual one.
Renamed on NgpThreadViewportState, NgpThreadState and the NgpThreadMessage call site, all @internal, so no public API change.

packages/ng-primitives/ai/src/thread-viewport/thread-viewport-state.ts
```diff
@@ -47,9 +60,18 @@ export const [
...
+        return;
+      }
+
+      scrollToBottom(behavior);
+    }
+
     function onScroll(): void {
       const { scrollHeight, scrollTop, clientHeight } = element.nativeElement;
-      const atBottom = scrollHeight - scrollTop <= clientHeight;
+      const atBottom = scrollHeight - scrollTop - clientHeight <= threshold();
```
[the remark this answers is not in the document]
pauldvlp: Confirmed and fixed. threshold() is only read inside onScroll(), which runs from the scroll listener and the resize subscription, so lowering the input from 70 to 0 while 50px from the bottom left isAtBottom stale at true and a streamed update still scrolled.
Recomputing through onScroll() is enough, scrollTop hasn't moved, so lastScrollTop >= scrollTop holds, the directional guard passes and the flag is re-derived against the new value:

```
// no scroll event fires when the threshold itself changes, so recompute against the new value
explicitEffect([threshold], () => onScroll());
```

Added should recompute the at-bottom state when the threshold changes: scroll to 450 (inside the 70px threshold), narrow the threshold to 0 without scrolling again, then stream and assert the viewport stays put. Confirmed it fails without the fix - 592 vs 450.

ashley-hunter (approved):
Looks good, thanks

PR #881, opened 2026-07-28
feat(ai): make the dictation language configurable and stop it overwriting user edits

pauldvlp:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

No issue — both parts were raised with you on Discord, where you were happy for them to land as one
PR.

Two problems in `NgpPromptComposerDictation`, both inside `initializeSpeechRecognition`.

### Dictation was overwriting the user's edits

The session is `continuous`, so it stays alive until the user clicks or presses Escape — the whole
of it is a window in which they can edit the field. But the prompt was snapshotted once in `onstart`
and never resynced, and `onresult` rebuilt the transcript from `results[0]` on every event. So the
primitive never appended, it **rewrote the whole field from a stale snapshot**. Anything typed or
deleted mid-session was discarded on the next result, and dictated text the user had removed came
back.

Repro: dictate `"hello"`, clear the input with dictation still running, say `"world"` — the field
read `"hello world"` instead of `"world"`.

The fix accumulates finalised speech incrementally, tracking how many results are already folded in
so none is counted twice, and re-derives the base when the prompt no longer matches what dictation
last wrote — so the user's edit wins instead of being overwritten. Ending a session resets that
state, and a stray result arriving after it can no longer rewrite the field.

I accumulate from a `finalisedCount` rather than from `event.resultIndex`: it is idempotent, so a
browser re-reporting an already finalised result cannot duplicate it.

### The dictation language was hardcoded

`recognition.lang` was fixed to `'en-US'`, and `recognition` is local to the factory, so there was
no way to reach it from outside. This adds an `ngpPromptComposerDictationLang` input _and_ a
`provideAiConfig` global config, as discussed. When neither is set the page's own language is used —
the document's `lang` attribute, then the browser's — falling back to `en-US` only if the page says
nothing.

The language is resolved per session rather than at construction, so a value bound after the button
is created applies to the next dictation.

### Test mock

`MockSpeechRecognition` only ever emitted a single result, so the suite could not express a
continuous session at all — the bug above was invisible to it. It now models the real API: an
interim result is refined and finalised at the index it already occupies, and each new phrase takes
the next one. All 12 existing dictation tests pass against it unchanged.

Seven tests are added. I confirmed the two that target the bug fail on `next`.

- [ ] Yes
- [x] No

No API is removed or changed — the input and the config are additive. Two behaviour changes worth
naming, both intended: edits made while dictation runs are now preserved instead of being
overwritten, and an application that relied on dictation always transcribing as `en-US` while
serving a page with a different `lang` will now follow the page. Setting
`ngpPromptComposerDictationLang="en-US"` restores the old behaviour.

---
## Summary by cubic
Make the dictation language configurable and prevent dictation from overwriting or restoring deleted text. Adds a global config and input; resolves language per session with fallbacks.

- **New Features**
  - Configure language via `ngpPromptComposerDictationLanguage` or global `provideAiConfig({ dictationLanguage })`; resolved per session with fallback: document `lang` → browser → `en-US`.
  - Exported `NgpAiConfig` and `provideAiConfig` from `ng-primitives/ai`; docs updated with a Global Configuration section.

- **Bug Fixes**
  - Preserve user edits by accumulating only new final results and rebasing when the prompt diverges; deleted interim text no longer reappears; edits made before the first result are kept; new sessions start from the current prompt and late results after stop are ignored.
  - Test mock now models continuous sessions; new tests cover edits (including before-first-result), interim deletion, and language resolution.

<sup>Written for commit fb674400c1b641a818b4e56d71a3071da8896c2c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ng-primitives/ng-primitives/pull/881?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added configurable dictation language for the AI assistant, with both per-component selection and application-wide defaults.
  * Dictation now supports a `language` input (including the `ngpPromptComposerDictationLanguage` alias) and resolves language with fallbacks.

* **Bug Fixes**
  * Improved dictation session handling to preserve user edits during active dictation.
  * Ensures interim/final transcript accumulation is consistent across start/stop cycles.

* **Documentation**
  * Added a “Global Configuration” section with a TypeScript example and API reference for setting default AI assistant options.

### DIFF
--- a/packages/ng-primitives/ai/src/config/ai-config.ts
+++ b/packages/ng-primitives/ai/src/config/ai-config.ts
@@ -1,0 +1,39 @@
+import { InjectionToken, Provider, inject } from '@angular/core';
+
+export interface NgpAiConfig {
+  /**
+   * The BCP 47 language tag dictation transcribes in, e.g. `en-US` or `es-ES`.
+   * When undefined the page's own language is used: the document's `lang` attribute,
+   * falling back to the browser's language.
+   * @default undefined
+   */
+  dictationLang: string | undefined;
+}
+
+export const defaultAiConfig: NgpAiConfig = {
+  dictationLang: undefined,
+};
+
+export const NgpAiConfigToken = new InjectionToken<NgpAiConfig>('NgpAiConfigToken');
+
+/**
+ * Provide the default Ai configuration
+ * @param config The Ai configuration
+ * @returns The provider
+ */
+export function provideAiConfig(config: Partial<NgpAiConfig>): Provider[] {
+  return [
+    {
+      provide: NgpAiConfigToken,
+      useValue: { ...defaultAiConfig, ...config },
+    },
+  ];
+}
+
+/**
+ * Inject the Ai configuration
+ * @returns The global Ai configuration
+ */
+export function injectAiConfig(): NgpAiConfig {
+  return inject(NgpAiConfigToken, { optional: true }) ?? defaultAiConfig;
+}
--- a/packages/ng-primitives/ai/src/config/ai-config.ts
+++ b/packages/ng-primitives/ai/src/config/ai-config.ts
@@ -7,11 +7,11 @@ export interface NgpAiConfig {
    * falling back to the browser's language.
    * @default undefined
    */
-  dictationLang: string | undefined;
+  dictationLanguage: string | undefined;
 }
 
 export const defaultAiConfig: NgpAiConfig = {
-  dictationLang: undefined,
+  dictationLanguage: undefined,
 };
 
 export const NgpAiConfigToken = new InjectionToken<NgpAiConfig>('NgpAiConfigToken');
--- a/packages/ng-primitives/ai/src/prompt-composer-dictation/prompt-composer-dictation-state.ts
+++ b/packages/ng-primitives/ai/src/prompt-composer-dictation/prompt-composer-dictation-state.ts
@@ -11,6 +11,9 @@ import {
 } from 'ng-primitives/state';
 import { injectPromptComposerState } from '../prompt-composer/prompt-composer-state';
 
+/** Used only when neither the consumer nor the page states a language. */
+const DEFAULT_DICTATION_LANG = 'en-US';
+
 export interface NgpPromptComposerDictationState {
   /**
    * Whether dictation is currently active.
@@ -23,6 +26,10 @@ export interface NgpPromptComposerDictationProps {
    * Whether the dictation button should be disabled.
    */
   readonly disabled?: Signal<boolean>;
+  /**
+   * The BCP 47 language tag dictation transcribes in. Falls back to the page's own language.
+   */
+  readonly lang?: Signal<string | undefined>;
 }
 
 export const [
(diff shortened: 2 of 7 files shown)


packages/ng-primitives/ai/src/config/ai-config.ts
```diff
@@ -0,0 +1,39 @@
+import { InjectionToken, Provider, inject } from '@angular/core';
+
+export interface NgpAiConfig {
+  /**
+   * The BCP 47 language tag dictation transcribes in, e.g. `en-US` or `es-ES`.
+   * When undefined the page's own language is used: the document's `lang` attribute,
+   * falling back to the browser's language.
+   * @default undefined
+   */
+  dictationLang: string | undefined;
```
ashley-hunter: Could we possibly use language instead of lang everywhere? Just for consistency, we have tended not to shorten or abbreviate words in our APIs, it is a bit more verbose but I think it’s better than mixing.

packages/ng-primitives/ai/src/prompt-composer-dictation/prompt-composer-dictation-state.ts
```diff
@@ -98,45 +114,85 @@ export const [
...
+        }
+
         let interimTranscript = '';
-        let finalTranscript = '';
 
-        // Process all results
-        for (let i = 0; i < event.results.length; i++) {
+        // Fold in only the results not already accounted for. Re-reading the list from zero would
+        // re-apply every phrase of the session on top of a base that may have moved on.
+        for (let i = finalisedCount; i < event.results.length; i++) {
```
ashley-hunter, line 149: I think here, if we dictate the word “hello” and then remove it while dictation is still active, it could be added back when the interim result is finalised. Since manual edits should take priority, could we add a test for this case and make sure deleted interim text is not restored?

ashley-hunter (changes requested):
Thanks, looks good! Looking into this a bit further I think I might have found one additional edge case I hadn’t previously covered

packages/ng-primitives/ai/src/config/ai-config.ts
pauldvlp: Sure! Piece of cake ☺️

packages/ng-primitives/ai/src/prompt-composer-dictation/prompt-composer-dictation-state.ts
pauldvlp: ☺️ 👍🏻

ashley-hunter (approved):
Great! Thank you!

PR #919, opened 2026-08-12
fix(number-field): treat non-finite values and bounds as unset

### LINKED ISSUE #917: NumberField: a NaN value binding bypasses setValue's NaN guard — renders "NaN" and disables both steppers
### Affected Primitive

Number

### Version

latest

## Summary

  `ngpNumberField` rejects `NaN` in `setValue`, but a `NaN` arriving through the
  `ngpNumberFieldValue` binding never passes through `setValue` — it lands in state
  unchecked. The result is the literal text `NaN` in the input and both steppers
  permanently disabled.

  `NaN` is a valid `number`, so this is not a typing mistake on the consumer's side. It
  turns up naturally from `Number(x)` where `x` is an optional field that hasn't been
  filled in yet.

## Reproduction

  StackBlitz: <your link>

  ```ts
  @Component({
    template: `
      <div [(ngpNumberFieldValue)]="value" [ngpNumberFieldMin]="0" [ngpNumberFieldMax]="100" ngpNumberField>
        <button ngpNumberFieldDecrement aria-label="Decrement">−</button>
        <input ngpNumberFieldInput />
        <button ngpNumberFieldIncrement aria-label="Increment">+</button>
      </div>
    `,
  })
  export class AppComponent {
    // realistic origin: Number(someOptionalField) before the user has entered anything
    value = signal(Number(undefined)); // NaN
  }

  Expected: the field renders empty (placeholder visible) and the steppers work,
  seeding from min the way an initially-null field does.

  Actual: the input shows NaN, the placeholder is suppressed, and both + and −
  carry data-disabled — so the field cannot be recovered through the UI at all except by
  typing over it.

### Reproduction

https://unzdju1u.stackblitz.io

rikeshchhunju-sketch:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

Closes #917

`setValue` rejects `NaN`, but a `NaN` arriving through the `ngpNumberFieldValue` binding never passed through it — it landed in `controlledState` unchecked. The input rendered the literal text `NaN`, the placeholder was suppressed, and both steppers carried `data-disabled` (`canIncrement` is `NaN < max` = `false`, `canDecrement` is `NaN > min` = `false`), so the field could not be recovered through the UI. `NaN` is a legitimate `number` and turns up naturally from `Number(someOptionalField)`.

While tracing it, two neighbouring holes in the same boundary turned up:

- `clampAndStep` runs `Math.max(min(), val)` **after** `setValue`'s guard, so a `NaN` `min`/`max` binding manufactures a `NaN` from a perfectly valid value and pushes it through both the internal setter **and** `valueChange`. Fixing only the value binding would have made this worse — the input would render empty while the consumer's signal silently received `NaN`.
- The guard was `isNaN`, not `!Number.isFinite`, so `±Infinity` survived `clampAndStep` intact under the default `max = Infinity`.

The fix states one rule at the state boundary: **stored value is always finite or `null`, and a `NaN` bound to a numeric option means unbound.**

### `number-field-state.ts`

- The value resolved by `controlledState` is normalised: non-finite → `null`. Normalising *after* resolution (rather than wrapping the input signal) covers the `value` binding, the `defaultValue` binding and `setDefaultValue` in one place, and structurally cannot disturb `controlledState`'s `undefined` sentinel — `isControlled` is latched before the new `computed` ever runs, so an absent binding still means uncontrolled.
- `min` / `max` / `step` / `largeStep` fall back to their defaults when bound `NaN`, via a small `defaultIfNaN` helper.
- `setValue`'s guard widened from `isNaN` to `!Number.isFinite`. Together with normalised bounds this makes the invariant airtight: `clampAndStep(finite)` is always finite, so no non-finite value can reach state by any path.
- The skip-if-unchanged check compares against the *stored* value rather than the normalised one, so a parent holding a non-finite value is still notified when a commit resolves it to `null`.

All normalisation lives in the state factory rather than the directive's input `transform`, since `ngpNumberField` is public API and can be composed directly — a consumer calling `ngpNumberField({ value: mySignal })` never sees a directive transform.

### `number-field-input-state.ts`

The display-sync effect re-implemented `formatDisplayValue()` inline; it now calls it. No defensive guard was added in the display layer — the invariant belongs to the factory, and a guard here would have covered only one of the four places the input writes its value while masking a regression from the DOM-based tests.

### Behaviour notes for review

- A bound `NaN` renders empty, while `setValue(NaN)` is still silently ignored. A binding declares the field's whole value so it must render *something*, and empty is the only honest rendering of "not a number"; an imperative call is a requested transition, and refusing it preserves the existing guarantee rather than letting one bad `Number(...)` wipe a good value.
- With `NaN` bound one-way, state reads `null` while the parent still holds `NaN`, and nothing is emitted to repair it. This matches existing behaviour — `[ngpNumberFieldValue]="150"` with `max="100"` is likewise not clamped or written back. Any stepper press or commit emits a real number.
- `[ngpNumberFieldStep]="NaN"` changes from "stepping silently no-ops" to "steps by 1". The old behaviour was an accidental fall-through of `isFinite(step())` in `clampAndStep`.
- `NaN` `min`/`max` → `∓Infinity` keeps `aria-valuemin`/`aria-valuemax` absent, which is already what the `isFinite`-guarded bindings do with `NaN` today. Only the value pipeline changes.

### Tests

17 tests added to `number-field-primitive.test.ts` — a new `non-finite bindings` block plus `setValue(Infinity)` alongside the existing `setValue(NaN)` case. 14 of them fail without this change; the three that don't are the two controlled/uncontrolled latching regression guards (which must pass either way) and the `NaN` step case, which already fell back to `1` by accident.

Coverage includes the reported repro end-to-end (a two-way binding starting as `Number(undefined)` recovering to `1` on the first `+`), `Infinity`, a non-numeric string through the input transform, `NaN` `defaultValue`, and explicit guards that an `undefined` binding stays uncontrolled while `null` stays controlled-and-empty.

### Docs

The `value` input's JSDoc gained "A non-finite value (`NaN`, `±Infinity`) is treated as empty", which flows into the generated API table via `<api-reference-props>` — no markdown change needed.

Full suite passes (2649 tests, no type errors); lint and prettier clean.

- [ ] Yes
- [x] No

---
## Summary by cubic
Treats non‑finite values and bounds in `ngpNumberField` as unset and keeps pagination CVAs valid on null writes. Previously, a bound `NaN` rendered literally and disabled steppers; inverted infinite bounds could emit `±Infinity`; pagination `writeValue(null)` left an invalid `data-page`.

- Normalizes after `controlledState`: non‑finite `value`/`defaultValue` → `null`; non‑finite `min`/`max` → unset; non‑finite `step`/`largeStep` → `1`/`10`; removes step finiteness guards now made unreachable.
- Enforces a finite‑or‑null invariant: `setValue` rejects non‑finite args and overflowed clamp results; “skip if unchanged” compares against the stored value so normalization still emits meaningful changes.
- Input fixes: shared formatter eliminates “NaN”; Home/End act only with finite bounds; omit `aria-valuemin/max` when unbounded; `inputmode` and the character filter derive from normalized `min`/`step`.
- Pagination: `writeValue(value: number | null)` now coerces `null` to page 1 in the reusable component and generated schematic, avoiding stale `data-page`.
- Notable behavior change: `[ngpNumberFieldStep]="NaN"` steps by `1` instead of no‑op; `setValue(±Infinity)` is ignored; `step="0"` performs no step.

<sup>Written for commit 5225672f38fe3369326b81f6a98559d53d5c37f2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ng-primitives/ng-primitives/pull/919?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved number field handling for invalid and non-finite values, including `NaN`, positive and negative infinity, and invalid numeric strings.
  * Invalid values now resolve safely to empty or default values where appropriate.
  * Prevented invalid values and overflow results from being accepted.
  * Pagination now resets to page 1 when a form supplies a null value.
  * Improved display synchronisation when number field values or settings change.

* **Documentation**
  * Clarified handling of non-finite number field values and configuration options.

### DIFF
--- a/apps/components/src/app/pages/reusable-components/pagination/pagination.ts
+++ b/apps/components/src/app/pages/reusable-components/pagination/pagination.ts
@@ -176,9 +176,11 @@ export class Pagination implements ControlValueAccessor {
   }
 
   /** Write a new value to the control */
-  writeValue(value: number): void {
+  writeValue(value: number | null): void {
+    // Angular writes null on reset and before the initial value lands; page is always
+    // a number, so fall back to the first page rather than passing null through.
     // writing a value from the model must not re-emit through onChange
-    this.state().setPage(value, { emit: false });
+    this.state().setPage(value ?? 1, { emit: false });
   }
 
   /** Register a callback to be called when the value changes */
--- a/packages/ng-primitives/number-field/src/number-field-input/number-field-input-state.ts
+++ b/packages/ng-primitives/number-field/src/number-field-input/number-field-input-state.ts
@@ -118,8 +118,8 @@ export const [
 
     // Sync input display value when the number field value changes
     // (programmatically, via stepping, or on commit)
-    explicitEffect([() => numberField().value()], ([value]) => {
-      elementRef.nativeElement.value = value !== null ? String(value) : '';
+    explicitEffect([() => numberField().value()], () => {
+      elementRef.nativeElement.value = formatDisplayValue();
     });
 
     listener(elementRef, 'focus', () => {
--- a/packages/ng-primitives/number-field/src/number-field-input/number-field-input-state.ts
+++ b/packages/ng-primitives/number-field/src/number-field-input/number-field-input-state.ts
@@ -167,7 +167,8 @@ export const [
 
       function getLargeStepMultiplier(): number {
         const s = numberField().step();
-        if (!Number.isFinite(s) || s <= 0) return 1;
+        // `step` is normalised to a finite value; guard only against dividing by zero
+        if (s <= 0) return 1;
         return numberField().largeStep() / s;
       }
 
--- a/packages/ng-primitives/number-field/src/number-field-input/number-field-input-state.ts
+++ b/packages/ng-primitives/number-field/src/number-field-input/number-field-input-state.ts
@@ -47,7 +47,7 @@ export const [
     const inputMode = computed(() => {
       const minVal = numberField().min();
       const stepVal = numberField().step();
-      const allowsNegative = !isFinite(minVal) || minVal < 0;
+      const allowsNegative = !Number.isFinite(minVal) || minVal < 0;
       const hasDecimals = stepVal % 1 !== 0;
 
       // Some mobile keyboards can't show both minus sign and decimal point
@@ -65,11 +65,11 @@ export const [
     attrBinding(elementRef, 'spellcheck', 'false');
     attrBinding(elementRef, 'aria-valuemin', () => {
       const min = numberField().min();
-      return isFinite(min) ? min.toString() : null;
+      return Number.isFinite(min) ? min.toString() : null;
     });
     attrBinding(elementRef, 'aria-valuemax', () => {
       const max = numberField().max();
-      return isFinite(max) ? max.toString() : null;
+      return Number.isFinite(max) ? max.toString() : null;
     });
     attrBinding(elementRef, 'aria-valuenow', () => numberField().value()?.toString() ?? null);
     attrBinding(elementRef, 'tabindex', () => tabindex().toString());
@@ -93,7 +93,7 @@ export const [
         numberField().setValue(null);
       } else {
         const parsed = parseFloat(trimmed);
-        if (!isNaN(parsed)) {
+        if (!Number.isNaN(parsed)) {
           numberField().setValue(parsed);
         }
       }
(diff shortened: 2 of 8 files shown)


ashley-hunter (approved):
Great catch, thank you

ashley-hunter:
@allcontributors add @rikeshchhunju-sketch for code

PR #922, opened 2026-08-13
feat(portal): let an open overlay follow a changing anchor

alexciesielski:
> [!NOTE]
> **Stacked on #921.** It targets `next` because the base branch lives on a fork and GitHub can't stack across forks, so the first two commits here are #921 and appear in this diff.
>
> Review the last two commits only — `feat(portal): let an open overlay follow a changing anchor` and `fix(portal): only rebuild anchor bindings when the reference moves`. Once #921 merges, rebasing collapses this to just those.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation changes

## Issue

Closes #

No linked issue. This follows directly from the limitation both reviewers independently raised on #921 — that `setAnchor()` and `[ngpMenuTriggerAnchor]` have no effect after the first open. Happy to move this to a discussion first if you'd rather settle the direction before reviewing code, since it touches the shared portal layer rather than one primitive.

`anchorElement` was read once, when the overlay was created, so the anchor inputs on the tooltip, popover and menu triggers only ever described *where the overlay opened*. Rebinding one afterwards did nothing.

It now also accepts a `Signal`, alongside the config's other reactive fields (`content`, `placement`, `position`, `context`), and tooltip, popover and menu pass theirs straight through.

### Why it isn't just a type change

`getPositionReference()` already re-read the anchor on every pass, so the positioning *maths* was nearly there. What wasn't were the four places that captured an **element** when they were built rather than reading one. Each now moves with the anchor:

| Capture | Why it has to follow |
| --- | --- |
| resize observer | watches the reference for a collapse to zero size, and hides the overlay when it goes |
| `CloseScrollStrategy` | resolves the reference's overflow ancestors in `enable()` |
| `autoUpdate` reference | closed over for the overlay's lifetime |
| registry entry | `register()` no-ops on a duplicate id, so the outside-press decision was frozen at open time |

That last one is what makes the current anchor count as inside for dismissal and the one it replaced count as outside — so exactly one element is inside at a time.

### Not a breaking change

`anchorElement` stays assignable from a plain element on both `NgpOverlayConfig` and `NgpOverlayEntry`, which are public. A plain element still means a fixed anchor, so existing `createOverlay` callers are unaffected. `overlay-anchor.test-d.ts` asserts the widening rather than leaving it to whichever caller happens to exercise it.

### Motivation

One menu shared across hundreds of interactive tokens — a virtualized JSON tree, also rendered as a grid cell renderer across many rows — where a trigger per token is far too much machinery. #921 alone doesn't help there: a static anchor gives nothing that moving a zero-size `position: fixed` span doesn't already give. Re-pointing a live anchor does, and deletes the workaround.

## Tests

Four tests in `menu-anchor.test.ts` covering the behaviours that only hold with a live anchor: an open menu moves to the new anchor; the new anchor is inside; the **previous** anchor becomes outside; unsetting the anchor while open falls back to the trigger. Reverting the portal change fails exactly those four and leaves #921's four passing.

Plus `overlay-anchor.test-d.ts` for the type compatibility above.

## Notes for review

Two things worth a second opinion:

**`handleAnchorChange` bails when the resolved element is unchanged.** Without that guard it disposed and re-created `autoUpdate` under a live overlay whenever a `controlled` input notified without the element actually differing, which made `menu-trigger-group-cooldown` fail intermittently — roughly 2 runs in 3, against 3 clean runs on #921's branch. With the guard, three consecutive full runs are clean. Flagging it because it was a genuine intermittent failure I introduced, not a pre-existing flake.

**`fromResizeEvent` is now passed an explicit `injector`.** An anchor change reaches it from an effect, where there is no injection context, and it calls `inject(PLATFORM_ID)`. This is equivalent to the previous implicit behaviour — `createOverlay` already runs construction inside `runInInjectionContext(config.injector, …)` — it just also works off the effect path.

- [ ] Yes
- [x] No

---
## Summary by cubic
Open overlays now follow a changing anchor and treat the current anchor as inside for outside‑press. Previously the anchor was captured on open; now `anchorElement` can be a signal so menus, popovers, and tooltips re‑anchor live without closing, and re‑anchoring avoids double position computation.

- `NgpOverlay` accepts `anchorElement` as an element or signal, tracks it in an internal signal, and updates all anchor‑bound bindings when the reference actually changes: resize observer (now created with the provided injector), floating‑ui `autoUpdate` reference, `CloseScrollStrategy` overflow ancestors, and the outside‑press registry (which now reads a live signal). `updateConfig({ anchorElement })` can swap between an element and a signal; anchor removal/zero‑size closes the overlay. Position is computed once when rebinding; teardown windows no longer leak listeners.
- `NgpMenuTrigger`, `NgpPopoverTrigger`, and `NgpTooltipTrigger` forward anchor signals and expose `setAnchor(element|null)` to claim anchors immediately. Docs include a dynamic‑anchor example and press timing guidance; tests cover live re‑anchoring, inside/outside behavior, press ordering, node survival, and type widening.

**Rollout**
- No breaking change; passing a plain element still yields a fixed anchor.
- If you re‑anchor during a press, call `setAnchor` on `pointerdown`. Setting the input or using `click` may dismiss instead of move.

<sup>Written for commit 93079cf1f05cb26f032a5c2798e3a8654495c5a0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/ng-primitives/ng-primitives/pull/922?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Menus, popovers and tooltips can now be positioned against custom anchor elements.
  - Changing the anchor while an overlay is open automatically repositions it and updates its dismissal boundary.
  - Menus support dynamically selecting targets while preserving consistent press, outside-click and keyboard behaviour.

- **Documentation**
  - Expanded custom-anchor guidance and examples, including keyboard accessibility and press interactions.

- **Tests**
  - Added coverage for dynamic repositioning, fallback behaviour, anchor replacement and dismissal handling.

### DIFF
--- a/apps/documentation/src/app/examples/menu/menu-dynamic-anchor.example.ts
+++ b/apps/documentation/src/app/examples/menu/menu-dynamic-anchor.example.ts
@@ -1,0 +1,189 @@
+import { Component, signal } from '@angular/core';
+import { NgpButton } from 'ng-primitives/button';
+import { NgpMenu, NgpMenuItem, NgpMenuTrigger } from 'ng-primitives/menu';
+
+@Component({
+  selector: 'app-menu-dynamic-anchor',
+  imports: [NgpButton, NgpMenu, NgpMenuTrigger, NgpMenuItem],
+  template: `
+    <div class="stack">
+      <div class="tokens">
+        @for (token of tokens; track token) {
+          <button
+            class="token"
+            #element
+            [class.selected]="anchor() === element"
+            (click)="claim(trigger, element)"
+            (pointerdown)="claim(trigger, element)"
+            type="button"
+          >
+            {{ token }}
+          </button>
+        }
+      </div>
+
+      <!-- Below the tokens, so the menu it opens never covers the row it moves across. -->
+      <button
+        class="open"
+        #trigger="ngpMenuTrigger"
+        [ngpMenuTrigger]="menu"
+        ngpMenuTriggerPlacement="bottom-start"
+        ngpMenuTriggerOffset="8"
+        ngpButton
+      >
+        Format value
+      </button>
+    </div>
+
+    <ng-template #menu>
+      <div ngpMenu>
+        <button ngpMenuItem>Copy</button>
+        <button ngpMenuItem>Copy path</button>
+        <button ngpMenuItem>Filter on value</button>
+      </div>
+    </ng-template>
+  `,
+  styles: `
+    :host {
+      display: contents;
+    }
+
+    .stack {
+      display: flex;
+      flex-direction: column;
+      gap: 1rem;
+      align-items: flex-start;
+    }
+
+    .tokens {
+      display: flex;
+      flex-wrap: wrap;
+      gap: 0.5rem;
+    }
+
+    .open {
+      height: 2.125rem;
+      padding: 0 0.625rem;
+      border: none;
+      border-radius: 0.625rem;
+      background-color: var(--ngp-background);
+      box-shadow:
+        inset 0 0 0 1px var(--ngp-border),
+        0 1px 2px 0 rgb(0 0 0 / 0.04);
+      font-size: 0.875rem;
+      font-weight: 510;
+      letter-spacing: -0.006em;
+      color: var(--ngp-text-primary);
+      transition: background-color 150ms cubic-bezier(0.4, 0, 0.2, 1);
+      outline: none;
+    }
+
+    .open[data-hover] {
+      background-color: var(--ngp-background-hover);
+    }
+
+    .open[data-press] {
+      background-color: var(--ngp-background-active);
+    }
+
+    .open[data-focus-visible] {
+      outline: 2px solid var(--ngp-focus-ring);
+      outline-offset: 1px;
+    }
+
+    .token {
+      height: 2.125rem;
+      padding: 0 0.625rem;
+      border: none;
+      border-radius: 0.5rem;
+      background-color: var(--ngp-background);
+      box-shadow: inset 0 0 0 1px var(--ngp-border);
+      font-family: ui-monospace, monospace;
+      font-size: 0.8125rem;
+      letter-spacing: -0.006em;
+      color: var(--ngp-text-secondary);
+      cursor: pointer;
+      transition:
+        background-color 150ms cubic-bezier(0.4, 0, 0.2, 1),
+        box-shadow 150ms cubic-bezier(0.4, 0, 0.2, 1);
+      outline: none;
+    }
+
+    .token:hover {
+      background-color: var(--ngp-background-hover);
+    }
+
+    .token:active {
+      background-color: var(--ngp-background-active);
+    }
+
+    .token:focus-visible {
+      outline: 2px solid var(--ngp-focus-ring);
+      outline-offset: 1px;
+    }
+
+    .token.selected {
+      box-shadow: inset 0 0 0 1.5px var(--ngp-primary);
+      color: var(--ngp-text-primary);
+      font-weight: 510;
+    }
+
+    [ngpMenu] {
+      position: fixed;
+      display: flex;
+      flex-direction: column;
+      width: max-content;
+      background: var(--ngp-background);
+      border: 1px solid var(--ngp-border);
+      box-shadow: var(--ngp-shadow-lg);
+      border-radius: 0.625rem;
+      padding: 0.25rem;
+      outline: none;
+      transform-origin: var(--ngp-menu-transform-origin);
+    }
+
+    [ngpMenuItem] {
+      padding: 0.4375rem 0.75rem;
+      border: none;
+      background: none;
+      cursor: pointer;
+      transition: background-color 0.15s ease;
+      border-radius: 0.375rem;
+      min-width: 140px;
+      text-align: start;
+      outline: none;
+      font-size: 0.875rem;
+      font-weight: 510;
+      letter-spacing: -0.006em;
+      color: var(--ngp-text-primary);
+    }
+
+    [ngpMenuItem][data-hover],
+    [ngpMenuItem][data-focus-visible] {
+      background-color: var(--ngp-background-hover);
+    }
+
+    [ngpMenuItem][data-press] {
+      background-color: var(--ngp-background-active);
+    }
+  `,
+})
+export default class MenuDynamicAnchorExample {
+  /** Drives the highlight only - the menu is re-anchored through `setAnchor`. */
+  readonly anchor = signal<HTMLElement | null>(null);
+
+  readonly tokens = ['"id"', '"name"', '"createdAt"', '"tags"'];
+
+  /**
+   * Called on `pointerdown` so the anchor is claimed before the capture-phase `mouseup`
+   * that decides whether the press landed outside - on `click` alone the press would
+   * dismiss the menu instead of moving it. `setAnchor` rather than the input, because an
+   * input only reaches the trigger on the next change detection pass and a fast tap can
+   * outrun it. The `click` binding keeps the targets usable without a pointer, where it
+   * claims the anchor for the next open rather than moving a menu already on screen.
+   */
+  claim(trigger: NgpMenuTrigger, element: HTMLElement): void {
+    trigger.setAnchor(element);
+    this.anchor.set(element);
+  }
+}
--- a/apps/documentation/src/app/examples/menu/menu-dynamic-anchor.tailwind.example.ts
+++ b/apps/documentation/src/app/examples/menu/menu-dynamic-anchor.tailwind.example.ts
@@ -1,0 +1,87 @@
+import { Component, signal } from '@angular/core';
+import { NgpButton } from 'ng-primitives/button';
+import { NgpMenu, NgpMenuItem, NgpMenuTrigger } from 'ng-primitives/menu';
+
+@Component({
+  selector: 'app-menu-dynamic-anchor-tailwind',
+  imports: [NgpButton, NgpMenu, NgpMenuTrigger, NgpMenuItem],
+  template: `
+    <div class="flex flex-col items-start gap-4">
+      <div class="flex flex-wrap gap-2">
+        @for (token of tokens; track token) {
+          <button
+            class="h-[2.125rem] cursor-pointer rounded-lg border-none bg-white px-2.5 font-mono text-[0.8125rem] tracking-[-0.006em] text-gray-600 outline-hidden transition-colors duration-150 hover:bg-gray-100 focus-visible:outline-2 focus-visible:outline-offset-1 focus-visible:outline-blue-500 active:bg-gray-200 dark:bg-transparent dark:text-gray-400 dark:hover:bg-white/10 dark:focus-visible:outline-blue-400 dark:active:bg-white/20"
+            #element
+            [class]="
+              anchor() === element
+                ? 'font-[510] text-gray-900 ring-2 ring-[#f01e2b] dark:text-gray-100 dark:ring-[#ff4651]'
+                : 'ring-1 ring-black/5 dark:ring-white/10'
+            "
+            (click)="claim(trigger, element)"
+            (pointerdown)="claim(trigger, element)"
+            type="button"
+          >
+            {{ token }}
+          </button>
+        }
+      </div>
+
+      <!-- Below the tokens, so the menu it opens never covers the row it moves across. -->
+      <button
+        class="h-[2.125rem] rounded-[0.625rem] border-none bg-white px-2.5 text-sm font-[510] tracking-[-0.006em] text-gray-900 shadow-sm ring-1 ring-black/5 outline-hidden transition-colors duration-150 data-focus-visible:outline-2 data-focus-visible:outline-offset-1 data-focus-visible:outline-blue-500 data-hover:bg-gray-100 data-press:bg-gray-200 dark:bg-transparent dark:text-gray-100 dark:ring-white/10 dark:data-focus-visible:outline-blue-400 dark:data-hover:bg-white/10 dark:data-press:bg-white/20"
+        #trigger="ngpMenuTrigger"
+        [ngpMenuTrigger]="menu"
+        ngpMenuTriggerPlacement="bottom-start"
+        ngpMenuTriggerOffset="8"
+        ngpButton
+      >
+        Format value
+      </button>
+    </div>
+
+    <ng-template #menu>
+      <div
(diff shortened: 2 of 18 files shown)


alexciesielski:
Downstream verification came back green on a 0.128.7 backport of this branch — all four moving-anchor behaviours confirmed against a real consumer (one menu shared across hundreds of JSON-tree tokens rendered in grid cells), and no regressions across their select, charts, menu, json-tree, tooltip, clipboard and popover suites.

Two findings from that pass are now in this PR.

**Event ordering is a real trap, so it's documented and tested.** Outside-press dismissal is decided on a capture-phase `mouseup` (`overlay-registry.ts`), which runs *before* a bubbling `click`. A consumer re-anchoring one menu across many elements who claims the new anchor in a `click` handler is therefore too late: at dismissal time the anchor still points at the previously pressed element, the press reads as outside, and the gesture meant to move the menu tears it down instead. Nothing wrong with the overlay — but it is not discoverable, and every one-menu-many-targets consumer will hit it.

`6a78c2b` documents it on the anchor input and `setAnchor` for menu, tooltip and popover, and on the menu docs page, then pins it down from both sides:

- claimed on `mousedown` → the menu stays open and moves to the new anchor
- claimed on `click` → the menu dismisses

The second is the control. Without it the first proves nothing: a harness where nothing dismisses looks identical to a pass.

**The tests dispatch the full gesture deliberately.** `pointerdown` → `mousedown` → `mouseup` → `click`, with a change detection pass between the button phases because a real press delivers them in separate tasks and a template-bound anchor only reaches the directive on the next pass. `userEvent.click` collapses exactly the ordering these tests exist to check — downstream had a re-anchoring test that passed with the bug present for that reason.

Reverting the portal change fails the `mousedown` case and leaves the `click` control passing, so it is load-bearing rather than decorative.

**One thing I want to be straight about:** the `menu-trigger-group-cooldown` fix in `346126a` is corroborated only by that suite going from ~2-in-3 failures to three consecutive clean full runs. Downstream has no cooldown or trigger-group coverage, so their green runs show no regression but do not exercise the `autoUpdate` teardown path that broke. If you'd like a deterministic test for it rather than absence-of-flake, say so and I'll add one — it needs a seam into the overlay internals, so I didn't invent one unprompted.

alexciesielski:
More downstream evidence, and one test improvement taken from it (`bc62ff3`).

**A second, independent consumer had built the same workaround.** Alongside the JSON-tree case that motivated this, the same team found an unrelated logs-grid context menu in their workspace doing the identical thing — one reused menu, a zero-size `position: fixed` element moved onto the clicked cell — written by different people who did not know the other existed. Its workaround for the ordering problem was different and worse: it lets the dismissal happen and reopens on the next animation frame, so today that grid tears the overlay down and rebuilds it on every cell-to-cell switch.

Two teams, two codebases, two independently invented workarounds, same wall. That is the argument for the docs note being in the anchor API rather than in each consumer's test file — the failure mode (menu vanishes on what should be a reposition) is identical every time, and it is not discoverable from the outside.

**Node identity is now asserted, not just position.** Their strongest suggestion, and they were right: closing and reopening at the new anchor satisfies a position assertion exactly as well as moving does. Both re-anchor tests now hold the overlay element across the change and assert it is the same node — which is precisely what distinguishes this from the rAF reopen above. These are regression guards rather than discriminators: they pass today either way, and exist so that a future change that starts tearing down cannot pass quietly.

**The change deleted API downstream, not just code.** Their menu wrapper had grown an injectable whose only purpose was to smuggle the `NgpOverlay` instance from the menu *content* back to the *trigger*, plus a public `reposition()` wrapping `updatePosition()`. Both are now deleted along with the workarounds.

They were explicit that this is an observation, not a request, and I am relaying it the same way: the overlay is not reachable from the trigger and `updatePosition()` is not exposed on the trigger state, so re-measuring after moving an anchor by hand required injector plumbing. With a live anchor none of that is needed, so there is nothing to fix here — it is a data point about where the surface is thin, for whatever it is worth later. I have not changed anything in that direction.

Unchanged from my last comment: the `menu-trigger-group-cooldown` fix still rests only on absence-of-flake, and their suites have no cooldown coverage to corroborate it. Their workspace-wide sweep (132 projects, ~17k tests) is still running; every targeted suite is green, but I would not treat the full sweep as confirmed until it reports.

alexciesielski:
Rebased onto `next` now that #921 has merged, so this is down to the portal change alone. Both review findings are fixed in `e1c02de` — I verified each against the code rather than taking them at face value, and both were real.

**P1 — close-scroll strategy rebuilt during teardown.** Correct, and the window is wider than it looks. `destroyOverlay()` clears the portal and disables the scroll strategy at the top, but `isOpen.set(false)` only runs at the very end, after the exit animation. So for the whole animation `isOpen()` is `true` while the strategy is already a `NoopScrollStrategy` — an anchor change there enabled a fresh `CloseScrollStrategy` that nothing disables again. The live portal, not `isOpen`, now decides whether there is anything to move, which covers the positioning rebuild in the same window too.

**P2 — `updateConfig()` ignored a replacement anchor.** Also correct, and a regression I introduced: before this PR every read went through `this.config.anchorElement`, so `updateConfig({ anchorElement })` worked. Normalising once in the constructor silently broke it — accepted and then ignored, which is worse than rejecting it. The anchor now lives in a signal of its own that `updateConfig()` writes to, so it also handles swapping an element for a signal, or vice versa.

**Docs accessibility.** Both bots flagged the `<span>`, and they're right. It's a `<button>` now, bound on **both** `mousedown` and `click`: `mousedown` is what beats the dismissal for pointer input, and `click` is what makes it reachable from the keyboard — keyboard activation raises `click` with no preceding `mousedown` and no outside press to lose the race against, so binding both is correct rather than redundant.

**On the tests for P2 — worth flagging.** My first version passed *without* the fix. Default placement is `top`, so the overlay sits above whatever it follows, and asserting "above the trigger" was true whether or not the anchor applied. Rewritten to compare against the anchor's own box, which the two cases cannot both satisfy: now 2/2 pass with the fix and 2/2 fail without it. Same failure mode the downstream team hit twice, and it only surfaced because I ran the revert.

Gate on the final state: build clean, lint 0 errors, 2694 browser + 53 node tests, no type errors.

Still outstanding and unchanged: the `menu-trigger-group-cooldown` fix in `f815a73` rests only on absence-of-flake (2-in-3 failures before, three consecutive clean runs after). Happy to add a deterministic test if you want one — it needs a seam into the overlay internals, so I haven't invented one unasked.

alexciesielski:
Went back over this branch specifically for cost, since `explicitEffect([this.anchorElement], …)` is now registered on every overlay rather than conditionally like the `placement` and `position` ones. Found one real problem, which is fixed in 017a278.

### Every re-anchor positioned twice

`setupPositioning()` calls floating-ui's `autoUpdate`, which invokes `update()` immediately as it binds. `handleAnchorChange` then called `updatePosition()` on top of that, so a single anchor change ran two full `computePosition` passes — each with the flip/shift/size middleware, so several layout reads — and two `portal.detectChanges()` calls.

`handleAnchorChange` now returns once it has rebuilt the binding, and `updatePosition()` stays as the fallback for the case where there is no positioning binding to rebuild. Suite unchanged at 2694/2694.

### The unconditional effect itself

For the record, in case the registration looked heavier than it is:

- Overlays are created lazily on first `show()`, so the effect count tracks *opened* overlays, not rendered triggers. A page with hundreds of tooltip triggers registers none until one opens.
- With no anchor configured, `anchorSource` is written once in the constructor and `controlled()` wraps an input that never changes, so `anchorElement` never notifies and the effect is never dirtied after its first run.
- That first run can't do any work either — the portal is always null at construction, so it gets no further than the resize monitor.
- `referenceElement` is read on every `computePosition` (so every scroll tick, and every frame under `trackPosition`), but it's a memoized `computed`; the `isSignal()` branch only executes when the anchor actually changes.

It can't be made conditional the way `placement` is without giving up `updateConfig({ anchorElement })`, since that's how an anchor can arrive after construction.

### What a re-anchor still costs

Rebuilding `autoUpdate` walks `getOverflowAncestors()` for both elements and attaches scroll and resize listeners to each, plus a ResizeObserver and an IntersectionObserver — and floating-ui captures the reference element at call time with no way to swap it, so the rebuild is inherent rather than something I can optimise away. Add the ResizeObserver swap for trigger-width tracking, and the `CloseScrollStrategy` rebuild under `scrollBehavior: 'close'`.

That's the right shape for a press-driven anchor, which is the use case this exists for, and re-anchoring within the same element is already free via the identity guard. It would be the wrong shape for an anchor driven by `mousemove` — that would want coalescing, and I haven't added any.

### One retention note

The overlay holds `monitoredElement` — the last anchor — for as long as the instance lives, and instances survive hide. In a virtualized list where rows are recycled, that keeps the last-anchored element's detached subtree reachable until the next re-anchor or trigger destroy. Bounded to one element, so I've left it alone, but it is a new class of retention: previously the only element held was the trigger.

### Also in this push

- 462453e types the test's `TemplateRef` against `NgpOverlayTemplateContext`. It only failed under `tsconfig.vitest.json`, which neither the build nor lint covers, so the suite stayed green over it — generics are erased, so it was never going to fail at runtime. Back to the same error count as `next`.

ashley-hunter:
Reviewed this in depth and pushed the result to your branch. Rebased onto current `next` first (clean, no conflicts), so #921 is no longer in the diff and this is now just the anchor work.

The core change is right, and the four captures you identified in the description are the correct four. Most of what follows is about the guidance around it rather than the mechanism.

## The main finding: `mousedown` was necessary but not sufficient

Claiming the anchor before the capture-phase `mouseup` fixes the **ordering**. It does nothing about the **timing**. `ngpMenuTriggerAnchor` is an input, so a new value only reaches the directive on the next change detection pass, and a press does not always leave one.

I only found this by building the runnable example and driving it in Chrome. Measured against it:

| press held | result |
| --- | --- |
| 0ms | menu **dismissed** |
| 8ms and longer | menu moved |

A human press is 50-150ms, so it works by hand. A fast tap does not, and neither does a frame the browser is busy rendering the hundreds of targets a shared menu exists for, which is exactly the case this feature is for.

Your `pressSequence` helper hides it, because it puts `fixture.detectChanges()` between the button phases. The comment on it even describes the assumption ("an anchor bound in the template only reaches the directive on the next pass") - it just reads as test plumbing rather than as a constraint on the API. Dispatching the same press with nothing in between reproduces the dismissal in a unit test.

**The fix I pushed** is `setAnchor()` on the menu, popover and tooltip triggers. The states already had the setter, the directives just did not expose it, and it writes the signal the dismissal check reads, so there is no pass to wait for. Both sides are pinned now: the input losing a single-frame press, and `setAnchor` surviving one. Deferring `setAnchor` by a single macrotask fails the latter, which is the whole of what it buys.

## Test coverage

Deleting the resize-observer and close-scroll-strategy rebuilds from `handleAnchorChange` left the entire 2700-test suite green. Positioning re-reads the anchor every pass, so it keeps working alone and masks the rest, and those two were the captures nothing exercised. Added coverage from both sides, so a rebuild that never happens and one wired to the wrong element are told apart:

- the previous anchor stops being watched for a collapse, and the new one starts
- the close-scroll strategy stops firing on the old ancestors and starts on the new
- an anchor removed from the DOM closes the overlay (the virtualized-row case)
- the tooltip and popover pass-throughs, whose one-line change to a signal anchor had no coverage at all

Separately, the pre-existing tooltip and popover anchor tests could not fail. Two things have to be true before "is it near the anchor" means anything and neither was: the panel needs an explicit size and position, because Floating UI writes `top`/`left` which a static element ignores, and the assertion has to wait, because `computePosition` resolves a task after the element is in the document. Both left the panel measuring 0,0. They passed with anchor support removed entirely. Fixed, and they now fail when it is.

## Also pushed

- **`refactor(portal)`**: split `boundReference` from `monitoredElement`. Whether the anchor moved was being answered by where the resize observer happened to be pointed. They agree today, but a second reason to move the observer would have silently stopped anchor changes being noticed, and the failure mode is "nothing rebuilds", which nothing reports. No behaviour change.
- **`docs(menu)`**: a runnable `menu-dynamic-anchor` example. The headline case existed only as an uncompiled snippet, so nothing checked the recommended pattern worked - and it did not. Building it also surfaced a layout where the open menu covered the row it was meant to move across.
- Guidance now separates order from timing and says which binding buys which, and recommends `pointerdown` over `mousedown` since it is the one event covering mouse, pen and touch alike.
- Corrected the keyboard claim. Activating anything outside an open menu closes it, which is pre-existing and reproducible against the static anchor example from #921, so the `click` binding claims the anchor for the next open rather than moving a menu already on screen. From the keyboard it is two steps.

## Things I suspected and then disproved

Recording these so nobody re-litigates them:

- **An anchor removed from the DOM does not strand the overlay.** `ResizeObserver` reports 0x0 on detach, so your existing collapse guard already closes it. Pinned with a test rather than changed.
- **`updateConfig` double-computing is not worth a branch.** Measured 5 computes against 3 for the signal path, and the 3 are floating-ui's own `autoUpdate` rebind rather than anything here. `updateConfig` also has zero callers in the library.
- **The hover bridge measuring the trigger rect is correct.** The pointer leaves the trigger, so a trigger to menu corridor is the right one even when the anchor is elsewhere.

Your `handleAnchorChange` guard is doing real work, by the way. I checked the lifecycle windows it has to survive - show delay pending, exit animation plus `cancelDestruction`, `keepMounted` hidden, pending `closeTimeout` - and the anchor is re-read whenever anything next binds to it in each of them.

## Follow-ups, not this PR

- The live/frozen split in the overlay config is sharper now. `anchorElement` joins `content`, `placement`, `context` and `keepMounted` as live, while `offset`, `flip`, `shift`, `container`, `scrollBehaviour` and the delays are still snapshotted at first open. "The anchor is live" invites "so is the offset, right?".
- One extra `effect` and `computed` per overlay instance, including the majority with no anchor. It cannot be gated on `isSignal(config.anchorElement)` because `updateConfig` can swap a signal in later, which a test asserts. Verified zero behavioural delta: the reference is seeded during construction so the first effect run is a guaranteed no-op.

## Verification

3x full browser suite (2710 tests, no type errors), node suite, lint across 7 projects, all 5 builds, prettier clean. Mutation-checked the new tests so they are not decorative. Menu examples driven in a real browser, including the pre-existing `menu-custom-anchor` to confirm it still anchors to its chevron.

Thanks for this one, and for the detail in the description - the capture table in particular made it quick to work out what to attack.

ashley-hunter (approved):
All good, thanks!

